PR #1, opened 2015-06-30
Added order fulfillment auditing script

clintonb:
This script should be run periodically (every 15-30 minutes) to ensure that orders are being fulfilled.

XCOM-423

@rlucioni @Nickersoft @jimabramson @wedaly

### DIFF
--- a/auditing/order_fulfillment.py
+++ b/auditing/order_fulfillment.py
@@ -3,13 +3,12 @@
 """ Verify all completed orders have active enrollments. """
 from contextlib import closing
 from datetime import timedelta, datetime
-from itertools import groupby
 import logging
 import os
 import sys
 
 import MySQLdb
-from MySQLdb.cursors import DictCursor
+import pandas
 
 logger = logging.getLogger(__name__)
 
@@ -17,7 +16,7 @@ logger = logging.getLogger(__name__)
 def setup_logging():
     """ Setup logging.
 
-    Add support for console logging.
+    Add support for console logging, and avoid truncation by pandas.
     """
     msg_format = '%(asctime)s - %(levelname)s - %(message)s'
     ch = logging.StreamHandler()
@@ -25,6 +24,10 @@ def setup_logging():
     ch.setLevel(logging.DEBUG)
     logger.addHandler(ch)
     logger.setLevel(logging.DEBUG)
+
+    pandas.set_option('display.max_rows', 500)
+    pandas.set_option('display.max_columns', 500)
+    pandas.set_option('display.width', 1000)
 
 
 ecommerce_db = MySQLdb.connect(host=os.environ['ECOMMERCE_DB_HOST'],
@@ -38,14 +41,13 @@ edxapp_db = MySQLdb.connect(host=os.environ['EDXAPP_DB_HOST'],
                             db=os.environ['EDXAPP_DB_NAME'])
 
 # This is the number of minutes to look back when retrieving orders (e.g. all orders in the last 15 minutes).
-ORDER_WINDOW_START_TIME = int(os.environ.get('ORDER_WINDOW_START_TIME', 15))
+ORDER_WINDOW_START_TIME = int(os.environ.get('ORDER_WINDOW_START_TIME', 20))
 
 
 def get_orders(date_placed):
     """ Retrieve all completed orders placed after the given date. """
     sql = """SELECT
     u.username
-    , u.email
     , o.number
     , o.date_placed
     , o.total_excl_tax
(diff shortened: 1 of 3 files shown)


auditing/order_fulfillment.py
```diff
@@ -0,0 +1,166 @@
...
+    JOIN catalogue_productattributevalue pav ON (pav.product_id = ol.product_id)
+    JOIN catalogue_productattribute pa ON (pa.id = pav.attribute_id AND pa.code = 'course_key')
+    JOIN ( SELECT pav2.product_id, pav2.value_text
+    FROM catalogue_productattributevalue pav2
+    JOIN catalogue_productattribute pa2 ON (pa2.id = pav2.attribute_id AND pa2.code = 'certificate_type')
+    ) AS modes ON (modes.product_id = ol.product_id)
+    LEFT JOIN refund_refundline rl ON (rl.order_line_id = ol.id)
+WHERE
+    o.status = 'Complete'
+    AND o.date_placed > %s
```
jimabramson: i'm surprised quotes aren't needed around the date string, does the db adapter handle that in `execute()` ?

clintonb: Yes, quotes are added by the library to limit the possibility of SQL
injection.


-- 
Clinton Blackburn

auditing/order_fulfillment.py
```diff
@@ -0,0 +1,166 @@
...
+    , e.course_id
+    , e.mode
+    , e.is_active
+FROM
+    student_courseenrollment e
+    JOIN auth_user u ON (u.id = e.user_id)
+WHERE
+    u.username IN ({usernames})
+    AND e.course_id in ({course_ids});
+""".format(usernames=','.join('%s' for __ in usernames), course_ids=','.join('%s' for __ in course_ids))
```
rlucioni: Why not interpolate the usernames and course IDs directly? Does this pattern allow you to use `execute()` to escape the arguments?

clintonb: I don't quite understand your question. Passing a single, quoted, string makes this bit of code vulnerable to SQL injection attacks. Creating multiple placeholders—one for each username and course ID—gets around this issue.

rlucioni: I may have inferred something from your answer to @jimabramson above that isn't true. I'm wondering why you wouldn't do the following:
```python
format(usernames=','.join(u for u in usernames), course_ids=','.join(c for c in course_ids))
```
Why is this attack vector closed when you substitute the usernames and course IDs later on by calling `execute()`?

clintonb: Your sample code is invalid. A simple join won't work. These are strings that would need to be wrapped in quotes, and the quotes are the problem. `MySQLdb` raises an exception if you attempt to pass quotes.

rlucioni: Makes sense now. Thanks.

jimabramson:
@clintonb how long is this taking to run against full-size data?  does it run against live databases or replicas?

clintonb:
@jimabramson The script is run against replicas, and takes less than one second to execute when looking at orders placed in the last 15 minutes. When looking at all orders for the lifetime of Oscar (not a likely/recommended use), execution takes about 1-2 minutes.

clintonb:
@jimabramson I updated this script to use [`pandas`](http://pandas.pydata.org/pandas-docs/stable/index.html) which is much more efficient than my horrible nested loops. Execution when looking at the previous 20 minutes' worth of orders is about 100ms. When comparing orders for the life of Oscar—33K orders, 41K enrollment—execution time is about 7s. Obviously, execution time depends on the amount of data being read from the DB (which is not that much at 1AM!); but, we should be able to use this for any future debugging purposes without worrying about execution time.

@mulby @brianhw @jab5569 I *just* learned how to use pandas (by which I mean, I learned how to find the right StackOverflow pages). Would one of you review this code, please?

rlucioni:
@clintonb :+1:, nice use of pandas.

clintonb:
@mulby @jimabramson Any feedback or approvals?

jimabramson:
@clintonb I don't know pandas very well either but this looks sane to me, and it's great that the performance is in a better place.   :+1:

mulby:
I realize this is already merged, but I'll take a pass at it anyway, I don't see anything glaring here... just a few nits / suggestions.

auditing/order_fulfillment.py
```diff
@@ -0,0 +1,162 @@
...
+    o.status = 'Complete'
+    AND o.date_placed > %s
+    AND rl.id IS NULL
+ORDER BY
+    u.username ASC
+    , pav.value_text ASC
+    , o.date_placed ASC;
+"""
+    date_placed = date_placed.strftime('%Y-%m-%d %H:%M:%S')
+    orders = pandas.read_sql_query(sql, ecommerce_db, index_col='number', params=(date_placed,),
```
mulby: I suspect you can omit `index_col` here and avoid the call to `reset_index` below. Does that work as expected?

auditing/order_fulfillment.py
```diff
@@ -0,0 +1,162 @@
...
+                                   parse_dates=('date_placed',))
+    orders.reset_index(level=0, inplace=True)
+    original_order_count = len(orders)
+
+    # Identify upgrades and remove the honor purchase.
+    # TODO We will need to do the same for credit.
+    grouped = orders.groupby(['username', 'course_id'])
+    for key, row_ids in grouped.groups.iteritems():
```
mulby: Consider just iterating over the keys and then using `grouped.get_group(key)` to grab a complete DataFrame for that group, which you can then sort in place etc.

auditing/order_fulfillment.py
```diff
@@ -0,0 +1,162 @@
...
+    """ Identify the orders that do not have corresponding enrollments. """
+    merged = pandas.merge(orders, enrollments, on=['username', 'course_id', 'mode'], how='left')
+    unfulfilled_orders = merged[merged.is_active.isnull()]
+    return unfulfilled_orders
+
+
+def run_audit():
+    date_placed = datetime.utcnow() - timedelta(minutes=ORDER_WINDOW_START_TIME)
+    orders = get_orders(date_placed)
+    usernames = pandas.unique(orders.username.ravel())
```
mulby: I typically just do: `orders.username.unique()`

Is that consequentially different from using `ravel()`?

PR #2, opened 2016-04-26
Several improvements to sailthru content load

PaulWattenberger:


sailthru_content/sailthru_content.py
```diff
@@ -0,0 +1,317 @@
...
+    """ Setup logging.
+
+    Add support for console logging, and avoid truncation by pandas.
+    """
+    msg_format = '%(asctime)s - %(levelname)s - %(message)s'
+    ch = logging.StreamHandler()
+    ch.setFormatter(logging.Formatter(msg_format))
+    ch.setLevel(logging.DEBUG)
+    logger.addHandler(ch)
+    logger.setLevel(logging.DEBUG)
```
cpennington: I'm not sure, but I think you could use https://docs.python.org/2/library/logging.html#logging.basicConfig to shorten this up a bit.

sailthru_content/sailthru_content.py
```diff
@@ -0,0 +1,317 @@
...
+            response = sc.api_get('content', {'items': 1000})
+
+            if not response.is_ok():
+                logger.error("Error code %d connecting to Sailthru content api: %s",
+                             response.json['error'], response.json['errormsg'])
+                return
+
+            for body in response.json['content']:
+                logger.info(body)
+            return
```
cpennington: This `return` means that you only run through the `while True` loop once. Is that just a temporary debugging measure?

cpennington:
A high-level comment: These scripts would likely be easier to read/review if `main` was split into some smaller more focused functions. I often have a function for specifying the argument parser, and then a function of each of the main script tasks (`list`, `clear` and `load` in this case).

cpennington:
It would be nice to have https://travis-ci.org/ run the tests automatically, since you've written them.

sailthru_content/sailthru_content.py
PaulWattenberger: Actually, the Sailthru content API doesn't have paging.  I'll get rid of the while True, it's confusing

sailthru_content/sailthru_content.py
PaulWattenberger: I was trying to be consistent with other scripts

requirements.txt
```diff
@@ -1,2 +1,3 @@
 MySQL-python==1.2.5
 pandas==0.16.2
+sailthru-client
```
clintonb: Add a version. Also, given the divergent natures of these scripts, each script directory should have its own requirements file. There's no reason you should install either of the existing requirements for this script.

clintonb: My comment was not addressed. :(

sailthru_content/sailthru_content.py
clintonb: As the writer of those other scripts, I don't recommend trying to be consistent with them. They could probably use some TLC.

requirements.txt
PaulWattenberger: Patience is a rare and valuable virtue, though not one I possess myself....

PR #3, opened 2016-04-29
SailThru user import script

brianfohl:
Please advise on how to best write unit tests for these. The import script itself only has one function, which will upload a file into SailThru's API. There is no way to "mock" this or mark the import as only a test, so the users will actually get added in SailThru. We could add them to a dev-only list in the Dev account, and then make a GET call to verify the users added, but it's arguable this approach falls outside the scope of unit testing; as the processing of the import happens on SailThru's side. There would be dependency there which is outside of our codebase.

It could be possible to write a simple unit test for the CSV splitter script, but this would require mocking up user data in CSV and including that with the test file in GitHub.

In my opinion, these steps seem unnecessary for scripts which have already been functionally tested and will only need to be run a few times at most.

It should be noted that the auditing script also found in this repository does not have unit tests.


cpennington:
Actually, you could mock the client, if need be. The `patch` function in the `mock` library can be used to override values at the module scope in other modules.

sailthru_import/mailchimp-export.py
```diff
@@ -0,0 +1,71 @@
+"""
+This script is deprecated, but can be used as a reference for MailChimp and SailThru API interaction.
```
cpennington: I don't think it makes much sense to check this in, if we're not planning on using it.

sailthru_import/sailthru_import.py
```diff
@@ -0,0 +1,87 @@
...
+                time.sleep(30)
+                response = client.api_get('job', {'job_id': job_id})
+                if response.get_body().get("status") == "completed":
+                    return
+        else:
+            error = response.get_error()
+            print ("Error: " + error.get_message())
+            print ("Status Code: " + str(response.get_status_code()))
+            print ("Error Code: " + str(error.get_error_code()))
+    except SailthruClientError as e:
```
cpennington: I'd actually just let this exception raise, rather than explicitly printing it.

sailthru_import/sailthru_import.py
```diff
@@ -0,0 +1,87 @@
...
+                print("waiting for import to complete...")
+                time.sleep(30)
+                response = client.api_get('job', {'job_id': job_id})
+                if response.get_body().get("status") == "completed":
+                    return
+        else:
+            error = response.get_error()
+            print ("Error: " + error.get_message())
+            print ("Status Code: " + str(response.get_status_code()))
+            print ("Error Code: " + str(error.get_error_code()))
```
cpennington: In this case, the script won't exit with a non-zero status code, which means that it can't easily be run as part of an automated process.

sailthru_import/sailthru_import.py
```diff
@@ -0,0 +1,87 @@
...
+    except SailthruClientError as e:
+        print ("Exception")
+        print (e)
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Sailthru user import script')
+    parser.add_argument('sailthru_key', help='Sailthru access key.')
+    parser.add_argument('sailthru_secret', help='Sailthru access secret.')
```
cpennington: I'd make these optional configuration parameters that by default read from `os.env`.

sailthru_import/sailthru_import.py
```diff
@@ -0,0 +1,87 @@
...
+    parser.add_argument('csv_file_name', help='CSV file name for import.')
```
cpennington: You can use `nargs` to all the user to pass multiple csvs.

sailthru_import/test-api-calls.py
```diff
@@ -0,0 +1,71 @@
+"""
+This script is deprecated, but can be used as a reference for SailThru API interaction.
```
cpennington: Again, not sure it's worth checking this in.

sailthru_import/test-api-calls.py
```diff
@@ -0,0 +1,71 @@
...
+    #     print ("Error: " + error.get_message())
+    #     print ("Status Code: " + str(response.get_status_code()))
+    #     print ("Error Code: " + str(error.get_error_code()))
+
+
+
+except SailthruClientError as e:
+    # Handle exceptions
+    print ("Exception")
+    print (e)
```
clintonb: Add a new line.

sailthru_import/test-api-calls.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+
+
+
+
+try:
+    user = get_user(email_id)
+
+    # response = sailthru_client.api_get("user", {"id": email_id_fake})
```
clintonb: Remove commented-out code.

clintonb:
Why not use `log.info()` instead of `print()`?

sailthru_import/sailthru_import.py
brianfohl: The code does allow multiple CSVs to be passed, but I'm not using nargs.

brianfohl:
Would you recommend I switch to log.info() instead of print? This is intended to be run as a command line script.

clintonb:
Yes. Depending on where this is run, the logs will go to Splunk, where we will have them if we need to debug.

brianfohl:
I've made the requested changes, aside from using Mock for tests. As I'm pretty new to Python, I've never used Mock before - can I set up time with either of you guys to sit down and talk through the use of Mock and how we can apply it here? I've read the docs (https://docs.python.org/dev/library/unittest.mock.html) but some guidance for this specific case would be extremely helpful and save time.

cpennington:
@brianfohl I can sit down with you to chat about that the next time we're both around our desks.

brianfohl:
Added test case using Mock. Ready for review.

sailthru_import/test/test_sailthru_import.py
```diff
@@ -0,0 +1,14 @@
...
+import mock
+from mock import patch
+
+
+@patch('sailthru.sailthru_client')
+class SailThruTestCase(unittest.TestCase):
+    def test_upload_csv_to_sailthru(self, mock_client):
+        mock_client.return_value.api_get.response.body.status.return_value = "completed"
+        mock_client.upload_csv_to_sailthru('test.csv', 'list name', 'bfohl@edx.org', mock_client)
+        mock_client.upload_csv_to_sailthru.assert_called_once()
```
cpennington: As we talked about: use `sailthru_import.upload_csv_to_sailthru`, and then `mock_client.api_post.assert_called_once`.

Also, `assert_called_once_with(...)` will allow you to validate the arguments that are passed to sailthru.

sailthru_import/test/test_sailthru_import.py
```diff
@@ -9,6 +9,7 @@
 @patch('sailthru.sailthru_client')
 class SailThruTestCase(unittest.TestCase):
     def test_upload_csv_to_sailthru(self, mock_client):
-        mock_client.return_value.api_get.response.body.status.return_value = "completed"
-        mock_client.upload_csv_to_sailthru('test.csv', 'list name', 'bfohl@edx.org', mock_client)
+        mock_client.api_get.return_value.get_body.return_value = {'status': 'completed'}
+        mock_client.api_post.return_value.is_ok.return_value = True
+        sailthru_import.upload_csv_to_sailthru('test.csv', 'list name', 'bfohl@edx.org', mock_client)
         mock_client.upload_csv_to_sailthru.assert_called_once()
```
cpennington: Since `upload_csv_to_sailthru` isn't a method on the sailthru client, I'm surprised if this passes.

sailthru_import/test/test_sailthru_import.py
```diff
@@ -0,0 +1,15 @@
...
+from mock import patch
+
+
+@patch('sailthru.sailthru_client')
+class SailThruTestCase(unittest.TestCase):
+    def test_upload_csv_to_sailthru(self, mock_client):
+        mock_client.api_get.return_value.get_body.return_value = {'status': 'completed'}
+        mock_client.api_post.return_value.is_ok.return_value = True
+        sailthru_import.upload_csv_to_sailthru('test.csv', 'list name', 'bfohl@edx.org', mock_client)
+        mock_client.upload_csv_to_sailthru.assert_called_once()
```
cpennington: Since upload_csv_to_sailthru isn't a method on the sailthru client, I'm surprised if this passes.

sailthru_import/test/test_sailthru_import.py
```diff
@@ -0,0 +1,15 @@
...
+
+import unittest
+import mock
+from mock import patch
+
+
+@patch('sailthru.sailthru_client')
+class SailThruTestCase(unittest.TestCase):
+    def test_upload_csv_to_sailthru(self, mock_client):
+        mock_client.api_get.return_value.get_body.return_value = {'status': 'completed'}
```
cpennington: You can just create the client in this method, by calling `mock_client = Mock(spec=SailthruClient)`. That way, the mock will raise an error if any method is called on it that's not available on the actual client.

sailthru_import/sailthru_import.py
```diff
@@ -0,0 +1,85 @@
...
+Example usage:
+    $ sailthru_import AJKHSDKJHSADKJSH-1 KSJLAFUHDSUIH-2 input.csv "Dev Test" "bfohl@edx.org"
+
+Upload file size limit is 100 MB. csv_splitter is used to break up the export file from MailChimp.
+
+csvcut (part of csvkit) is recommended for filtering out the necessary fields:
+    csvcut -c "Email Address","activated","Age","Country","date_joined",
+    "DSTOFF","EUID","Gender","GMTOFF","id","last_login","LEID","MEMBER_RATING",
+    "REGION","TIMEZONE","Username","year_of_birth"
+    ~/Downloads/members_export_107aacb22a.csv -e ISO-8859-1 > Downloads/members_export_filtered_2.csv
```
cpennington: We probably should look at making this part of the scripted process, so that we can just point this script at a mailchimp export and walk away.

PR #5, opened 2016-05-10
Add affiliate cookie tracking script from Google Tag Manager.

peter-fogg:
# [ECOM-4329](https://openedx.atlassian.net/browse/ECOM-4329)

@clintonb @bderusha Can you review? FYI @rock345


bderusha:
belated 👍

PR #6, opened 2016-05-17
Add stage version of affiliate cookie script.

peter-fogg:
FYI @bderusha @clintonb


clintonb:
Can we move the scripts to the `gtm_scripts` directory?

👍

bderusha:
👍

peter-fogg:
@clintonb Ah, that directory didn't exist when I first added the script. Done.

PR #8, opened 2016-05-23
Update affiliate cookie domain.

peter-fogg:
# [ECOM-4329](https://openedx.atlassian.net/browse/ECOM-4356)

FYI @clintonb @bderusha


clintonb:
👍

PR #13, opened 2016-09-14
Duplicate site_name in content vars for enroll template

PaulWattenberger:
Duplicate site_name in content vars for enroll template.  Enroll template needs site_name and only 'vars' values passed to purchase api in otto


PaulWattenberger:
@schenedx fyi

PR #14, opened 2016-09-21
ECOM-5312 - Update sailthru content library with both courses and programs

schenedx:
I am considering this PR to be ready for review.
I significantly refactored @PaulWattenberger's original script to synch only the courses.
Please review @edx/ecommerce


sailthru_content/services/catalog_api_service.py
```diff
@@ -0,0 +1,51 @@
...
+            response = self.api_client.courses().get(limit=500, offset=(page * 500))
+
+            results = response['results']
+
+            if response['next']:
+                page += 1
+            else:
+                page = -1
+
+            for course in results:
```
clintonb: `course_collection += results`

sailthru_content/services/sailthru_api_service.py
```diff
@@ -0,0 +1,97 @@
+import json
```
clintonb: Correct the import ordering:

1. Python
2. Third-party
3. Local

sailthru_content/fixups.csv
```diff
@@ -1,19 +1,8 @@
...
 course-v1:AdelaideX+HumBio101x+1T2016,var,course_end,2017-12-31
 course-v1:AdelaideX+HumBio101x+1T2016,,expire_date,2017-12-31
 course-v1:UPValenciaX+BSP101x+3T2015,var,course_end,2017-12-31
 course-v1:UPValenciaX+BSP101x+3T2015,,expire_date,2017-12-31
 course-v1:BerkeleyX+CS110x+2T2016,var,course_end,2017-12-31
-course-v1:BerkeleyX+CS110x+2T2016,,expire_date,2017-12-31
-course-v1:HKUSTx+EBA101x+3T2016,var,course_start,2016-09-13
-course-v1:Microsoft+DEV218x+3T2016,var,course_start,2016-09-19
\ No newline at end of file
+course-v1:BerkeleyX+CS110x+2T2016,,expire_date,2017-12-31
```
rlucioni: Why is this data hardcoded?

sailthru_content/services/catalog_api_service.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+logger = logging.getLogger()
+COURSES_PAGE_SIZE = 500
+PROGRAMS_PAGE_SIZE = 40
+
+
+class CatalogApiService(object):
+
+    _cached_programs = {}
```
rlucioni: Mutable class attributes are dangerous! If someone modifies this attribute by accessing it through a particular instance, the referenced object accessed by all other instances of the class will be changed.

This should be an instance attribute set during initialization. Also, don't call something cached if it's not really cached.

sailthru_content/services/sailthru_api_service.py
```diff
@@ -0,0 +1,98 @@
...
+                sailthru_content_list.append(body)
+
+        return sailthru_content_list
+
+    def _upload_batch_file(self, filepath, report_email=None):
+        """
+        Use Sailthru job API to upload all content as a batch job
+        """
+        logging.info("Uploading %s" % filepath)
+        # Start the upload job
```
rlucioni: This comment adds no value.

sailthru_content/services/sailthru_translation_service.py
```diff
@@ -0,0 +1,275 @@
...
+        """
+        if tag:
+            resp = tag.replace(' & ', '-').replace(',', '').replace('.', '').replace(' ', '-').replace('--', '-')
+            if tagtype:
+                resp = tagtype + '-' + resp
+            return resp
+        return ''
+
+    def _create_course_vars(self, course, course_run, url, site_name, program_dictionary=None):
+        """ Generate 'vars' section of Sailthru data
```
rlucioni: Nitpick: keep docstring format consistent with our other repos (i.e., put the closing quote on the same line, or put the docstring on its own line).

sailthru_content/services/sailthru_translation_service.py
```diff
@@ -0,0 +1,275 @@
...
+            program_dictionary,
+        )
+        sailthru_content['spider'] = 0
+
+        # perform any fixups
+        if self.fixups:
+            for row in self.fixups:
+                if row[0] == course_run['key']:
+                    if row[1] == 'var':
+                        logging.info('Changing var.%s to %s for %s', row[2], row[3], row[0])
```
rlucioni: Storing these fixups in a named tuple or class would make this way more readable than using numbered indices.

sailthru_content/services/sailthru_translation_service.py
```diff
@@ -0,0 +1,275 @@
...
+        sailthru_item = {
+            'title': program.get('title'),
+            'url': program.get('marketing_url'),
+            'images': {
+                'thumb': {
+                    'url': program_card_image or program_banner_image
+                }
+            },
+            'description': program.get('subtitle'),
+            'spider': 0,
```
rlucioni: What does this do? A comment about "spider" would actually be helpful.

sailthru_content/services/tests/test_sailthru_api_service.py
```diff
@@ -0,0 +1,133 @@
+#! /usr/bin/env python
```
rlucioni: This line does no harm, but it's not necessary. Same applies to all other files using it.

sailthru_content/sync_library.py
```diff
@@ -0,0 +1,224 @@
...
+        if args.type in ('courses', 'all'):
+            sailthru_items.extend(translation_service.translate_courses())
+
+        if args.type in ('programs', 'all'):
+            sailthru_items.extend(translation_service.translate_programs())
+
+        if args.command == 'upload':
+            sailthru_service.upload(sailthru_items, args.report_email)
+        else:
+            print (json.dumps(sailthru_items))
```
rlucioni: Is this script meant to be run with Python 2 or Python 3?

sailthru_content/sync_library.py
```diff
@@ -0,0 +1,224 @@
...
+
+    elif args.command == 'clear':
+        sailthru_service.clear()
+    else:
+        print ('here is the help printout')
```
rlucioni: This is not helpful.

sailthru_content/services/tests/test_sailthru_api_service.py
```diff
@@ -0,0 +1,133 @@
...
+import json
+import mock
+import unittest
+
+from factory import fuzzy
```
rlucioni: FactoryBoy now includes the `factory.Faker` class. Built-in fuzzers are deprecated in favor of their [Faker](https://factoryboy.readthedocs.io/en/latest/#realistic-random-values) equivalents.

sailthru_content/services/sailthru_translation_service.py
```diff
@@ -0,0 +1,275 @@
...
+        }
+
+        tags = self._get_tags_from_property(program.get('subjects'), 'name', 'subjects')
+        tags.extend(self._get_tags_from_property(program.get('authoring_organizations'), 'key', 'school'))
+        tags.append(self._convert_tag(program.get('type').lower(), program.get('title')))
+
+        if len(tags) > 0:
+            sailthru_item['tags'] = ", ".join(tags)
+
+        # worry about modified date (date) and expire_date
```
rlucioni: I'm worried. Now what? (Remove this.)

sailthru_content/services/sailthru_translation_service.py
```diff
@@ -0,0 +1,275 @@
...
+        return sailthru_item
+
+    def _select_course_run_for_program(self, course):
+        course_run_list = course.get('course_runs')
+        if not course_run_list:
+            return None
+
+        sorted_course_runs = sorted(
+            course_run_list,
+            key=lambda run: datetime.datetime.strptime(run.get('start'), "%Y-%m-%dT%H:%M:%SZ"))
```
rlucioni: Nitpick: Keep your quote levels consistent.

sailthru_content/services/sailthru_translation_service.py
```diff
@@ -0,0 +1,275 @@
...
+        sailthru_content_vars['course_id'] = course['key']
+        sailthru_content_vars['course_run_id'] = course_run['key']
+
+        # duplicate site_name in vars for purchase/enroll templates
+        if site_name:
+            sailthru_content_vars['site_name'] = site_name
+
+        # use enrollment_end for sailthru expire_date
+        if course_run.get('enrollment_end'):
+            sailthru_content_vars['enrollment_end'] = self._convert_date(course_run['enrollment_end'])
```
rlucioni: Instead of a giant block of if statements, change the `_convert_date` method so that it accepts a dictionary and a key, and handles adding a new value only if the key is present in `course_run`. That way you write one if statement instead of six, and can use a for loop to handle the keys.

sailthru_content/services/sailthru_translation_service.py
```diff
@@ -0,0 +1,275 @@
...
+        return ''
+
+    def _create_course_vars(self, course, course_run, url, site_name, program_dictionary=None):
+        """ Generate 'vars' section of Sailthru data
+        """
+        sailthru_content_vars = {}
+        sailthru_content_vars['course_run'] = True
+        sailthru_content_vars['marketing_url'] = url
+        sailthru_content_vars['course_id'] = course['key']
+        sailthru_content_vars['course_run_id'] = course_run['key']
```
rlucioni: Just define the dictionary as you want it at the outset:
```py
sailthru_content_vars = {
    'course_run': True,
    ...
}
```

sailthru_content/services/sailthru_translation_service.py
```diff
@@ -0,0 +1,275 @@
...
+        if iso_date is None:
+            return None
+        try:
+            if '.' in iso_date:
+                return datetime.datetime.strptime(iso_date, "%Y-%m-%dT%H:%M:%S.%fZ").strftime("%Y-%m-%d")
+            return datetime.datetime.strptime(iso_date, "%Y-%m-%dT%H:%M:%SZ").strftime("%Y-%m-%d")
+        except ValueError:
+            return None
+
+    def _convert_datetime(self, iso_date):
```
rlucioni: You can collapse this method into the one above. They share most of their code, with the parsing being the only difference.

sailthru_content/services/sailthru_api_service.py
```diff
@@ -0,0 +1,98 @@
...
+            os.unlink(tmp_file.name)
+
+    def clear(self):
+        while True:
+            response = self.sailthru_client.api_get('content', {'items': 4000})
+
+            if not response.is_ok():
+                logging.error(
+                    "Error code %d connecting to Sailthru content api: %s",
+                    response.json['error'],
```
rlucioni: Does the response returned by the client really provide an API for checking if the status code is 200, but nothing to get at the error?

sailthru_content/services/tests/test_sailthru_api_service.py
```diff
@@ -0,0 +1,133 @@
...
+
+from services.sailthru_api_service import SailthruApiService
+
+GLOBAL_COUNTER = 0
```
rlucioni: This doesn't appear to be used anywhere.

sailthru_content/fixups.csv
schenedx: This is a temporary override requested by marketing on some of the missing data in the course catalog api. It's a quick dirty way to fix a people problem

sailthru_content/services/catalog_api_service.py
schenedx: This is not a django application. This is a script. So in a script setting, the data saved into the attribute is "cached" for the next call. I can certainly think of another name for this.

sailthru_content/services/sailthru_api_service.py
schenedx: The `SailthruResponse` actually don't check http status code. The implementation of the response object is at https://github.com/sailthru/sailthru-python-client/blob/master/sailthru/sailthru_response.py#L19

sailthru_content/services/sailthru_translation_service.py
schenedx: Agree. However, the fixups are a temporary measure that I hope to remove in the near future. I don't think it would worth the effort to make this prettier now

sailthru_content/services/tests/test_sailthru_api_service.py
clintonb: I believe your comment was meant for @rlucioni. I agree that we should use `faker` if `fuzzy` has been deprecated. See [`fake.pystr()`](https://faker.readthedocs.io/en/latest/providers/faker.providers.python.html).

schenedx: @clintonb Sorry, you are right. My comment was directed to @rlucioni. But you are correct by pointing out `fuzzy` is getting deprecated. I'll make the change to use `faker` instead

schenedx:
@clintonb @rlucioni I believe I have address all your comments here. Please take another pass

sailthru_content/services/sailthru_translation_service.py
rlucioni: There are still unaddressed instances of this.

sailthru_content/services/sailthru_translation_service.py
rlucioni: Are these "fixups" data that should be corrected in the catalog?

sailthru_content/services/sailthru_translation_service.py
```diff
@@ -0,0 +1,259 @@
...
+            for tag_item in source_array:
+                tags.append(self._convert_tag(tag_prefix, tag_item.get(key)))
+
+        return tags
+
+    def translate_course_run(self, course_run, course, program_dictionary=None):
+        # get marketing url
+        url = course_run['marketing_url']
+        if not url:
+            url = course['marketing_url']
```
rlucioni: url = course_run.get('marketing_url', course['marketing_url'])

sailthru_content/sync_library.py
rlucioni: @schenedx a note (maybe in the README) about which version of the Python the script supports would be good to have.

sailthru_content/services/sailthru_translation_service.py
schenedx: @rlucioni Yes, these are. But they are not yet corrected in course catalog. The most likely reason is the disagreements between marketing team and the course team on these courses.

PR #16, opened 2016-09-28
[ECOM-5539] Add GTM script to store UTM querystring params in a cookie

bderusha:
@edx/ecommerce please review

### DIFF
--- a/gtm_scripts/utm_tracking/stage_utm_cookie.js
+++ b/gtm_scripts/utm_tracking/stage_utm_cookie.js
@@ -1,0 +1,44 @@
+<script type="text/javascript">
+    var iCookieLengthDays = 90; // Cookie length in days
+    var sCookieName = "stage.edx.utm"; // Name of the first party cookie to utilise for last click referrer de-duplication
+    var sSourceParameterName = "utm_source"; // The parameter used by networks and other marketing channels to tell you who drove the traffic
+    var sMediumParameterName = "utm_medium"; // The parameter to identify the type of referrer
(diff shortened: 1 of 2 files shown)


gtm_scripts/utm_tracking/stage_utm_cookie.js
```diff
@@ -0,0 +1,43 @@
...
+            var aQueryStringParts = aQueryStringPairs[i].split("=");
+            if (sParameterName.toLowerCase() == aQueryStringParts[0].toLowerCase()) {
+                return aQueryStringParts[1];
+            }
+        }
+    };
+
+    var _setCookie = function (sCookieName, sCookieContents, iCookieLength) {
+        var dCookieExpires = new Date();
+        dCookieExpires.setTime(dCookieExpires.getTime() + (iCookieLength * 24 * 60 * 60 * 1000));
```
schenedx: Can you put an variable for expression `24 * 60 * 60 * 1000`? Like `millisecondsInDay`?

rlucioni:
@bderusha this is pretty new to me, so forgive me if the answer is otherwise obvious: is there a way to reduce the duplication we've got here? Most of the code in the two files touched here is the same.

bderusha:
Unfortunately no.  These files are just to track the changes we are making directly in the GTM ui.  One for each environment

bderusha:
The code is running on stage.edx.org if you want to check out the cookie


https://stage.edx.org/?utm_source=github&utm_medium=pr&utm_campaign=ECOM-5539

PR #18, opened 2016-10-03
Update Homepage report to include micromasters

bderusha:
Add micromasters enrollment events to enrollment reports.
Add micromasters landing pages to page reports.

ECOM-5850

FYI @edx/ecommerce

### DIFF
--- a/google_analytics/ga_homepage_report.py
+++ b/google_analytics/ga_homepage_report.py
@@ -37,6 +37,7 @@ recipients = ["ojchang@edx.org"]
 
 import argparse
 
+# pip install --upgrade google-api-python-client
 from apiclient.discovery import build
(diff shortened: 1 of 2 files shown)


google_analytics/ga_homepage_report.py
```diff
@@ -37,6 +37,7 @@
 
 import argparse
 
+# pip install --upgrade google-api-python-client
```
clintonb: Remove.

clintonb: Actually, add this to `requirements.txt`.

bderusha: @clintonb merged this before github showed me your comments.  I'll add this to `requirements.txt`

PR #23, opened 2016-11-08
Script for pulling translation files from Transifex

rlucioni:
This script can be used to automatically pull translations from Transifex, commit, push, and merge them to their respective repos. The script handles the interactions with GitHub required to create and merge PRs.

ECOM-5658.

@edx/ecommerce, please review. I'm talking with the tools and IT teams about automating execution of this script. In the meantime, it is possible to run this locally; I did so to create https://github.com/edx/course-discovery/pull/427, https://github.com/edx/programs/pull/176, and https://github.com/edx/credentials/pull/119.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,11 +1,10 @@
 edx-rest-api-client==1.6.0
+factory_boy==2.7.0
 google-api-python-client==1.5.3
 MySQL-python==1.2.5
 numpy==1.11.1
 pandas==0.16.2
+responses==0.5.1
 sailthru-client==2.2.3
 vertica-python==0.6.7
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -1,11 +1,11 @@
 edx-rest-api-client==1.6.0
+edx-rest-api-client==1.6.0
```
schenedx: Why do we specify this library twice in this file?

rlucioni: @schenedx because it was so disorganized before that no one noticed. I'll remove the duplicate line. Going forward, scripts should specify their own requirements. There's no need to install all of these requirements to run one of the scripts in here.

PR #24, opened 2016-11-09
Update homepage report for featured card events

bderusha:
[ECOM-5971](https://openedx.atlassian.net/browse/ECOM-5971)

Add requirements.txt for dependencies
Create new report script

FYI @edx/ecommerce


google_analytics/ga_homepage_report.py
```diff
@@ -1,365 +1,534 @@
...
-worksheet.set_column('C:S', 15, percent_fmt)
-worksheet.write('A1', 'Top Subject Pages from the Homepage, Data from '+str(d1)+' to '+str(d2) , cell_format)
-
-worksheet = writer.sheets['CourseEnrolls']
-
-worksheet.conditional_format('B1:B1000', {'type': '3_color_scale'})
+from apiclient.discovery import build
+from oauth2client import client, file, tools
+import pandas
+from pandas import DataFrame, ExcelWriter
```
MatthewPiatetsky: you're using pandas in two different ways

google_analytics/ga_homepage_report.py
```diff
@@ -1,365 +1,534 @@
...
-worksheet.set_column('A:A', 100)
-worksheet.set_column('B:B', 20, comma_fmt)
-worksheet.write('A1', 'Course Enrollments from the Homepage, Data from '+str(d1)+' to '+str(d2) , cell_format)
-worksheet.autofilter('A3:B3')
-
-writer.save()
+################################################################
+#
+# Run this script by providing the following inputs
+#     python ga_homepage_report.py yyyy-mm-dd yyyy-mm-dd /path/to/report/destination/
```
MatthewPiatetsky: maybe we want to use the bash date api here?  
Then you could just copy and paste the command rather than filling in those values

google_analytics/ga_homepage_report.py
```diff
@@ -1,365 +1,534 @@
...
+#
+# The first argument is the start date
+# The second argument is the end date
+# The third argument is the location the report will be written to
+#
+###############################################################
+
+def get_service():
+    """Get a service that communicates to a Google API."""
+    api_name = 'analytics'
```
MatthewPiatetsky: should all these variables be CONSTANTS defined at the top of the file?

google_analytics/ga_homepage_report.py
```diff
@@ -1,365 +1,534 @@
...
+
+def featuredProgramCardDrivenProgramEnrollments(date):
+    program_enrollment_events = ','.join([
+        'ga:eventAction==edx.bi.user.course-details.enrolled-user.enroll-card',  # New MicroMaster
+        'ga:eventAction==edx.bi.user.course-details.enroll.enroll-card',  # New xSeries
+        'ga:eventAction==edx.bi.user.course-details.enroll.discovery-card',  # Old MicroMaster
+        'ga:eventAction==edx.bi.user.xseries-details.enroll.discovery-card',  # Old xSeries
+    ])
+
+    featured_program_card_click = ';'.join([
```
MatthewPiatetsky: why are some events joined by , and some by ;

google_analytics/ga_homepage_report.py
```diff
@@ -1,365 +1,534 @@
...
+    dataframe['enrollsPerClickByType'] = dataframe['uniqueEnrolls'] / total_clicks_by_type
+
+    return dataframe
+
+
+def strip_edx_page_title(page_title):
+    return page_title.replace(' | edX', '')
+
+
+def output_report(filename, total_homepage_views=None, total_course_card_clicks=None, total_program_card_clicks=None, featured_cards=None, homepage_subjects=None):
```
MatthewPiatetsky: maybe we want the output_report thing in a separate file?

google_analytics/ga_homepage_report.py
```diff
@@ -1,365 +1,534 @@
...
+    )
+
+
+    # Create the homepage courses worksheet
+    if featured_cards is not None:
+        featured_cards.to_excel(writer, index=False, sheet_name='Featured Card Report', startrow=18)
+        featured_cards_worksheet = writer.sheets['Featured Card Report']
+
+        # Set conditional format
+        # featured_cards_worksheet.conditional_format('F1:F1000', {'type': '3_color_scale'})
```
MatthewPiatetsky: commented code?

google_analytics/ga_homepage_report.py
bderusha: GA syntax uses , for OR operations and uses ; for AND operations

google_analytics/ga_homepage_report.py
```diff
@@ -1,365 +1,534 @@
...
+    # RETURN total events & unique events
+    # GROUPED BY date, label (course/program id), card index
+    # FOR anyone who clicked a certain featured card
+    # AND whose session sequence was
+    #     featured card click EVENTUALLY FOLLOWED BY
+    #     a set of enrollment events
+    service = get_service()
+
+    data = service.data().ga().get(
+        ids='ga:' + '86300562',
```
bderusha: Didn't change this as part of the refactor.  Assuming it's our GA id?

google_analytics/ga_homepage_report.py
```diff
@@ -1,365 +1,534 @@
...
+        # featured_cards_worksheet.write('B5', float(featured_cards['uniqueViews'].sum())/total_homepage_views, percent_fmt)
+        # featured_cards_worksheet.write('B6', int(featured_cards['uniqueEnrolls'].sum()), comma_fmt)
+        # featured_cards_worksheet.write('B7', float(featured_cards['uniqueEnrolls'].sum())/total_homepage_views, percent_fmt)
+
+        featured_cards_worksheet.write(
+            'A1',
+            'Homepage Course Enrollments, Data from {start} to {end}'.format(start=start_date, end=end_date),
+            cell_format
+        )
+        featured_cards_worksheet.write('A3', 'Overview', cell_format)
```
MatthewPiatetsky: rather than repeatedly using write, what if you generate a list and then used something like write_list

google_analytics/ga_homepage_report.py
bderusha: They are only ever used within this method.  I thought it more readable to keep them within the method

google_analytics/ga_homepage_report.py
bderusha: ?

google_analytics/ga_homepage_report.py
```diff
@@ -1,365 +1,534 @@
...
+        total_homepage_views=total_homepage_views,
+        total_course_card_clicks=total_course_card_clicks,
+        total_program_card_clicks=total_program_card_clicks,
+        featured_cards=featured_card_df,
+        homepage_subjects=subject_dataframe
+    )
+
+
+# parse script arguments
+start_date = datetime.strptime(sys.argv[1], '%Y-%m-%d').date()
```
MatthewPiatetsky: maybe do the argument parsing in a method?

google_analytics/ga_homepage_report.py
bderusha: One for accessing objects like `DataFrame` the other for calling pandas class methods like `pandas.merge`.  Again, thought it more readable to work that way.  Potentially more python to do `from pandas import merge as pandas_merge` and have one import line?

google_analytics/ga_homepage_report.py
bderusha: This is a script run by marketing which is being sunset.  I'd prefer to keep it all together.

google_analytics/ga_homepage_report.py
MatthewPiatetsky: that's what I was thinking. If you think just merge will be confusing you could alias it with import as x

google_analytics/ga_homepage_report.py
MatthewPiatetsky: For instance this in bash gives you today's date $(date +%Y-%m-%d) on the command line (though after seeing this again I realized these are two custom dates, not today's date so ignore me)

google_analytics/ga_homepage_report.py
MatthewPiatetsky: Well based on your comment that this is being sunsetted I don't know if we care about the code quality.
However, the value would be that you could define the things you're writing compactly in a couple of lines and only calling the statement fewer times would mean this block would be like 5 lines instead of 40

PR #25, opened 2016-11-10
Add option to pass a clone URL as an argument

rlucioni:
Facilitates writing individual Jenkins jobs for each repo.

ECOM-5658

@edx/ecommerce please review.

### DIFF
--- a/transifex/pull.py
+++ b/transifex/pull.py
@@ -5,7 +5,14 @@ commit, push, and merge them to their respective repos.
 To use, export an environment variable `GITHUB_ACCESS_TOKEN`. The token requires
 GitHub's "repo" scope.
 
-Run the script from the root of this repo: `python transifex/pull.py`.
+Run the script from the root of this repo. To use the repo clone URLs stored in
+settings.yaml, simply run:
+
+    python transifex/pull.py
+
+If you want to run the script for a specific repo, provide a clone URL as an argument:
+
+    python transifex/pull.py git@github.com:edx/course-discovery.git
 """
 import concurrent.futures
 from contextlib import contextmanager
@@ -15,6 +22,7 @@ import os
 from os.path import join, abspath, dirname
 import re
 import subprocess
+import sys
 import time
 
 from github import Github, GithubException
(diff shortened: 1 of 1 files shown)


transifex/pull.py
```diff
@@ -218,11 +226,16 @@ def pull(repo):
...
-
-    with concurrent.futures.ProcessPoolExecutor() as executor:
-        for clone_url in settings['repos']:
-            repo = Repo(clone_url)
-            executor.submit(pull, repo)
+    try:
+        clone_url = sys.argv[1]
+        repo = Repo(clone_url)
+        pull(repo)
+    except IndexError:
```
schenedx: Why use exception to do what the `if...else...` statements can do? Can't you just check if the `sys.argv[1]` exists?

transifex/pull.py
```diff
@@ -218,11 +226,16 @@ def pull(repo):
...
-    settings_file = join(abspath(dirname(__file__)), 'settings.yaml')
-    with open(settings_file) as f:
-        settings = yaml.load(f)
-
-    with concurrent.futures.ProcessPoolExecutor() as executor:
-        for clone_url in settings['repos']:
-            repo = Repo(clone_url)
-            executor.submit(pull, repo)
+    try:
+        clone_url = sys.argv[1]
```
schenedx: It maybe a good idea to printout to the console that you are cloning from the provided git repo URL

transifex/pull.py
```diff
@@ -218,11 +226,16 @@ def pull(repo):
...
+        repo = Repo(clone_url)
+        pull(repo)
+    except IndexError:
+        settings_file = join(abspath(dirname(__file__)), 'settings.yaml')
```
schenedx: Before here, it would be nice to print to console that you are clone the repo specified in the settings YML file.

transifex/pull.py
rlucioni: I guess it's a stylistic choice. This is what exception handlers are meant for. You could achieve the same effect with `len()` and a conditional, too. I think what I have here is pretty explicit.

transifex/pull.py
rlucioni: I'll go with your suggestion below. It's not necessary to have both messages. The absence of one implies the other.

transifex/pull.py
rlucioni: Updated.

transifex/pull.py
schenedx: I don't know about a stylistic choice where there are many discussion on using exceptions as flow controls. See http://softwareengineering.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why
I am on the side of we should use exceptions for real undesired cases.

transifex/pull.py
schenedx: Agree.

transifex/pull.py
rlucioni: This is an example of a common Python coding style: https://docs.python.org/3/glossary.html#term-eafp.

schenedx: I don't have a very strong opinion on this. Thanks for the discussions.

PR #26, opened 2016-11-16
Assorted refinements for the Transifex pull script

rlucioni:
A variety of issues only made themselves apparent when running the script in the context of a Jenkins job. This updates the script to 1) consume HTTPS clone URLs, 2) raise exceptions when shell commands exit with non-zero exit codes, 3) avoid cloning entire repo histories, and 4) always cleaning up the cloned repo, even if an exception is raised during the pull.

ECOM-5658

@edx/ecommerce please review. Related to https://github.com/edx-ops/edx-jenkins-job-dsl/pull/53.

### DIFF
--- a/transifex/pull.py
+++ b/transifex/pull.py
@@ -24,6 +24,7 @@ import re
 import subprocess
 import sys
 import time
+from urllib.parse import urlparse
 
 from github import Github, GithubException
 import yaml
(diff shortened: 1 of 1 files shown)


MatthewPiatetsky:
👍 
New to this system, but polling until the status checks complete seems a bit strange to me. 
Couldn't you have github trigger a script when the status checks complete rather than waiting?  
Not sure if this resource is unavailable while it is polling

rlucioni:
@MatthewPiatetsky that behavior was introduced prior to this PR and isn't changing here. That said, having GitHub trigger a script which either merges the PR or comments on it to get our attention seems like a reasonable future improvement.

PR #28, opened 2016-11-24
SKU of seats added in sailthru content

tasawernawaz:
@ahsan-ul-haq @schenedx Please review this.

### DIFF
--- a/sailthru/sailthru_content/services/sailthru_translation_service.py
+++ b/sailthru/sailthru_content/services/sailthru_translation_service.py
@@ -4,6 +4,8 @@ import logging
 
 from constants import GFA_COURSE_RUN_LIST
 
+REQUIRED_SEATS_TYPES = ['verified', 'professional']
+
 
 class SailthruTranslationService(object):
     """This is the service that converts the data in edX into what Sailthru can consume"""
@@ -97,6 +99,10 @@ class SailthruTranslationService(object):
 
         return tags
 
+    def _return_sku(self, seats):
+        """ Return dict of sku for a course run. """
+        return {seat['type']: seat['sku'] for seat in seats if seat['type'] in REQUIRED_SEATS_TYPES}
+
     def translate_course_run(self, course_run, course, program_dictionary=None):
         # get marketing url
         url = course_run.get('marketing_url', course.get('marketing_url'))
@@ -132,6 +138,10 @@ class SailthruTranslationService(object):
 
         if len(tags) > 0:
             sailthru_content['tags'] = ", ".join(tags)
+
+        seats = course_run.get('seats')
+        if seats:
+            sailthru_content['sku'] = self._return_sku(seats)
 
         sailthru_content['vars'] = self._create_course_vars(
             course,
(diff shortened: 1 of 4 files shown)


sailthru_content/services/sailthru_translation_service.py
```diff
@@ -97,6 +97,16 @@ def _get_tags_from_property(self, source_array, key, tag_prefix):
 
         return tags
 
+    def _return_sku(self, seats):
+        """ Return dict of sku for a course run. """
+        sku = {}
```
clintonb: Every seat type has a SKU. Why the limitations?

`return {seat['type']: seat['sku'] for seat in seats}`

tasawernawaz: It was a requirement that we only need `sku` of `verified` or `professional` seats.

sailthru_content/services/sailthru_translation_service.py
```diff
@@ -97,6 +97,16 @@ def _get_tags_from_property(self, source_array, key, tag_prefix):
...
+        for seat in seats:
+            seat_type = seat.get('type')
+            if seat_type in ['verified', 'professional']:
```
schenedx: Can you set the array `['verified', 'professional']` as a constant somewhere in this file?

sailthru_content/services/sailthru_translation_service.py
schenedx: You can actually use the one liner suggested by @clintonb with the if statement.

sailthru_content/services/sailthru_translation_service.py
```diff
@@ -133,6 +143,10 @@ def translate_course_run(self, course_run, course, program_dictionary=None):
         if len(tags) > 0:
             sailthru_content['tags'] = ", ".join(tags)
 
+        seats = course_run.get('seats')
+        if seats:
```
schenedx: I thought dictionary `{}` would still evaluate to `True`?

tasawernawaz: `seats` is a empty list `[ ]` and will evaluates `True` only when there is at least one `seat` type exists.

sailthru_content/services/sailthru_translation_service.py
tasawernawaz: added.

sailthru_content/services/tests/test_sailthru_translation_service.py
```diff
@@ -103,8 +104,13 @@ def _get_expected_sailthru_item(self, seat_type=None):
             'date': '2016-09-23'
         }
 
+        if seat_type in ['professional', 'verified']:
```
schenedx: You can just replace this list with the `sailthru_translation_service.REQUIRED_SEATS_TYPES`

sailthru_content/services/tests/test_sailthru_translation_service.py
```diff
@@ -70,7 +71,7 @@ def test_translate_programs(self):
         self.assertEqual(len(translated_programs[0]['vars']['course_runs']), 4)
 
     def _get_expected_sailthru_item(self, seat_type=None):
```
clintonb: This method doesn't return anything.

clintonb: It seems we have tests; but, they aren't actually being run by Travis. That should be fixed.

clintonb: Got it. The test is still broken.

PR #30, opened 2016-11-29
virtual env shell script

umar-nawaz:
To setup virtual env for Jenkins job. 
cc @schenedx

### DIFF
--- a/requirements/virtualenv.sh
+++ b/requirements/virtualenv.sh
@@ -1,0 +1,6 @@
+# Create the virtualenv and install requirements
+mkdir -p sailthru
+virtualenv sailthru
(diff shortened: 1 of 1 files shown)


clintonb:
Why is this necessary? Jenkins jobs already run in virtualenvs.

umar-nawaz:
I think we need to run `requirements` on Jenkins. Well, the other option can be directly hitting `pip install -r requirements/base.txt` from jenkins-dsl-job. I am unsure if we create `virtualenv` for this.

umar-nawaz:
Okay, We actually don't need this.
 thanks @schenedx for clearing the stuff.

PR #22, opened 2016-11-08
Provide course_type affiliation on courses Sailthru content library

ahsan-ul-haq:
ECOM-6267

Currently it is work in progress and I am working on writing the tests.
@schenedx @tasawernawaz

PR #34, opened 2016-12-06
Remove unsearchable courses from SailThru content feed

ahsan-ul-haq:
ECOM-6442

@schenedx please review

### DIFF
--- a/sailthru/requirements/base.txt
+++ b/sailthru/requirements/base.txt
@@ -1,2 +1,3 @@
+nose==1.3.7
 edx-rest-api-client==1.6.0
 sailthru-client==2.2.3
--- a/sailthru/sailthru_content/services/catalog_api_service.py
+++ b/sailthru/sailthru_content/services/catalog_api_service.py
@@ -48,6 +48,18 @@ class CatalogApiService(object):
         logger.debug('Get Courses called')
         return self._get_resource_from_api(self.api_client.courses(), COURSES_PAGE_SIZE, marketable=1)
 
+    def get_searchable_course_run_keys(self):
+        logger.debug('Get Searchable Course runs called')
+        searchable_course_keys = []
+        searchable_course_runs = self._get_resource_from_api(
+            self.api_client.course_runs(), COURSES_PAGE_SIZE, hidden=False, marketable=1
+        )
+        for course_run in searchable_course_runs:
+            if course_run.get('key'):
+                searchable_course_keys.append(course_run.get('key'))
+
+        return searchable_course_keys
+
     def get_program_dictionary(self):
         if not self._programs_dictionary:
             program_array = self._get_resource_from_api(
--- a/sailthru/sailthru_content/services/sailthru_translation_service.py
+++ b/sailthru/sailthru_content/services/sailthru_translation_service.py
@@ -1,4 +1,7 @@
 # -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+
 import datetime
 import logging
 
@@ -165,11 +168,13 @@ class SailthruTranslationService(object):
     def translate_courses(self):
         sailthru_item_array = []
         courses = self.data_service.get_courses()
+        serachable_run_keys = self.data_service.get_searchable_course_run_keys()
         program_dictionary = self.data_service.get_program_dictionary()
         for course in courses:
             course_runs = course.get('course_runs', [])
             for course_run in course_runs:
-                sailthru_item_array.append(self.translate_course_run(course_run, course, program_dictionary))
+                if course_run.get('key') in serachable_run_keys:
+                    sailthru_item_array.append(self.translate_course_run(course_run, course, program_dictionary))
 
         return sailthru_item_array
 
--- a/sailthru/sailthru_content/services/tests/catalog_api_test_mixins.py
+++ b/sailthru/sailthru_content/services/tests/catalog_api_test_mixins.py
@@ -7,7 +7,9 @@ import responses
 
 
 from sailthru.sailthru_content.services.catalog_api_service import CatalogApiService
-from sailthru.sailthru_content.services.tests.fixtures import SINGLE_COURSE_DATA, SINGLE_PROGRAM_DATA
+from sailthru.sailthru_content.services.tests.fixtures import (
+    SINGLE_COURSE_DATA, SINGLE_PROGRAM_DATA, SINGLE_SEARCHABLE_COURSE_DATA
+)
 
 
 class CatalogApiTestMixins(unittest.TestCase):
@@ -46,6 +48,15 @@ class CatalogApiTestMixins(unittest.TestCase):
         responses.add(
             responses.GET, self.api_url_root + '/programs/',
             body=json.dumps(SINGLE_PROGRAM_DATA),
+            status=200,
+            content_type='application/json',
+            match_querystring=False
+        )
+
+    def prepare_searchable_courses(self):
+        responses.add(
+            responses.GET, self.api_url_root + '/course_runs/',
+            body=json.dumps(SINGLE_SEARCHABLE_COURSE_DATA),
             status=200,
             content_type='application/json',
             match_querystring=False
--- a/sailthru/sailthru_content/services/tests/fixtures.py
+++ b/sailthru/sailthru_content/services/tests/fixtures.py
@@ -1,30 +1,32 @@
 # -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
 SINGLE_PROGRAM_DATA = {
     "count": 1,
     "results": [
         {
-            "uuid": u"281102cf-2546-4a25-8c04-985cb48c87cc",
-            "title": u"International Hospitality Management",
-            "subtitle": u"Earn a valuable credential in international hospitality and tourism management from one of the world’s leading programs to advance your career.",
-            "type": u"MicroMasters",
-            "status": u"active",
-            "marketing_slug": u"hkpolyux-international-hospitality-management",
-            "marketing_url": u"https://www.edx.org/micromasters/hkpolyux-international-hospitality-management",
+            "uuid": "281102cf-2546-4a25-8c04-985cb48c87cc",
+            "title": "International Hospitality Management",
+            "subtitle": "Earn a valuable credential in international hospitality and tourism management from one of the world’s leading programs to advance your career.",
+            "type": "MicroMasters",
+            "status": "active",
+            "marketing_slug": "hkpolyux-international-hospitality-management",
+            "marketing_url": "https://www.edx.org/micromasters/hkpolyux-international-hospitality-management",
             "courses": [
                 {
-                    "key": u"HKPolyUx+HTM541x",
-                    "title": u"International Hospitality Luxury Management",
-                    "short_description": u"Understand the power of a luxury brand and gain the business skills needed to provide premium goods and services within the hotel...",
-                    "full_description": u"<p>Learn strategies and skills for managing luxury brands within the hotel and tourism industry. You will analyze the essence of a luxury brand and learn marketing techniques and trends to better position your services. You will learn how luxury brands are created and how to properly communicate the value of your brand.</p>\n<p>Concepts and applications will be introduced through textbooks, selected journal articles and field visits.</p>",
-                    "level_type": u"Intermediate",
+                    "key": "HKPolyUx+HTM541x",
+                    "title": "International Hospitality Luxury Management",
+                    "short_description": "Understand the power of a luxury brand and gain the business skills needed to provide premium goods and services within the hotel...",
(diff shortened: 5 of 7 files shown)


schenedx:
@edx/ecommerce Please help review.
@clintonb I think we can use the search api for now to filter out what we can. When ECOM-6392 is implemented, we can create a new ticket to alter this script to use the better solution. Is that OK?

schenedx:
@ahsan-ul-haq I believe there are tests associated with this script and these components you updated. Can you update those unit tests? If you found some unit tests are broken, could you please spend the time fixing them? I want to see a unit test for this change.

schenedx:
@ahsan-ul-haq I realized that you didn't filter out programs based on the search returns. Can you please implement that to filter out programs like https://www.edx.org/xseries/edx-course-creator?
You can use the same /search/all/facets/ end point with a `content_type=program` parameter.

ahsan-ul-haq:
@schenedx made the suggested changes also updated tests and added new ones as well.

sailthru/sailthru_content/services/catalog_api_service.py
```diff
@@ -44,10 +45,45 @@ def _get_resource_from_api(self, api_endpoint, page_size, **kwargs):
 
         return results
 
+    def _get_resource_from_search_api(self, api_endpoint, **kwargs):
+        page = 1
+        response = api_endpoint.get(**kwargs)
```
schenedx: These statements are repeated with the statements in the while loop. I think you can find a way to set `next_page` to something that's truthful in the beginning and remove the repeated statements

sailthru/sailthru_content/services/sailthru_translation_service.py
```diff
@@ -69,7 +71,7 @@ def _create_course_vars(self, course, course_run, url, site_name, program_dictio
                 # multiple seat types the one with the higher precedence will be added
                 # in course_type
                 if not sailthru_content_vars.get('course_type', None) or \
-                        seat_priority.index(sailthru_content_vars['course_type']) < seat_priority.index(seat['type']):
+                        seat_priority.index(sailthru_content_vars['course_type']) > seat_priority.index(seat['type']):
```
ahsan-ul-haq: yes it is a bug after testing it seems like audit mode was having higher priority as compared to verified so I made this change.

schenedx: If this is a bug fix, please create another PR and fix this bug so we can merge that one in quickly and run this script against Sailthru prod with that bug fix. I want this fix to be in quicker, not being buried with this PR.

rlucioni:
@ahsan-ul-haq are your tests here being run by Travis? If not, please add a `.travis.yml` so that tests are actually run.

ahsan-ul-haq:
@rlucioni we already have a ticket related to this work https://openedx.atlassian.net/browse/ECOM-6471. So should I do that in this PR or we can do it in separate PR?

ahsan-ul-haq:
@rlucioni as simon is out on vaccations could you please review this?

ahsan-ul-haq:
@mikedikan please review

mikedikan:
Please rebase changes into single commit

rlucioni:
@ahsan-ul-haq as long as you're aware that the tests aren't being run and that they should be, it can be addressed separately. I didn't realize you had a story already. Since you do, introduce the .travis.yml in a separate commit.

sailthru/sailthru_content/services/tests/catalog_api_test_mixins.py
```diff
@@ -45,3 +53,12 @@ def prepare_get_programs(self):
             content_type='application/json',
             match_querystring=False
         )
+
+    def prepare_searchable_courses(self):
+        responses.add(
+            responses.GET, self.api_url_root + '/search/all/facets/',
```
mikedikan: I suspect that if you switch to the course run search endpoint: /api/v1/search/course_runs/ you will avoid the need to use the dictionary and the need to throw away programs entirely

sailthru/requirements/base.txt
```diff
@@ -1,2 +1,3 @@
+nose==1.3.7
```
rlucioni: @mikedikan it's the latest version of [nose](https://pypi.python.org/pypi/nose/) available. I would have suggested https://pypi.python.org/pypi/pytest, but that's besides the point.

rlucioni: @ahsan-ul-haq there's no need for a nested requirements file here. Please just use a requirements.txt.

sailthru/sailthru_content/services/catalog_api_service.py
```diff
@@ -44,10 +45,43 @@ def _get_resource_from_api(self, api_endpoint, page_size, **kwargs):
...
+        return results
+
     def get_courses(self):
         logger.debug('Get Courses called')
         return self._get_resource_from_api(self.api_client.courses(), COURSES_PAGE_SIZE, marketable=1)
 
+    def get_searchable_course_keys(self):
+        logger.debug('Get Searchable Courses called')
+        searchable_course_keys = []
+        searchable_courses = self._get_resource_from_search_api(self.api_client.search.all.facets())
```
ahsan-ul-haq: The idea here is that we want to get all the courses which are searchable in marketing site. I not sure whether course run end point only returns the searchable ones.

sailthru/sailthru_content/services/catalog_api_service.py
```diff
@@ -44,10 +45,43 @@ def _get_resource_from_api(self, api_endpoint, page_size, **kwargs):
...
+        for course in searchable_courses:
+            if course.get('key') and course.get('content_type') != 'program':
+                searchable_course_keys.append(course['key'])
+
+        return searchable_course_keys
+
+    def get_courses_dictionary(self):
```
ahsan-ul-haq: I am not sure about it.

sailthru/requirements/base.txt
ahsan-ul-haq: @rlucioni it was suggested by @clintonb here https://github.com/edx/ecommerce-scripts/pull/22#discussion_r87195920. So should I add requirements in a single or we should use nested requirements file?

sailthru/sailthru_content/services/catalog_api_service.py
ahsan-ul-haq: BTW I don't think the script kept-alive for more than a single call.

sailthru/requirements/base.txt
clintonb: Use a nested file. If I am running the Sailthru script, there is no reason I should need to install pandas, needed by the audit script. The same goes for needing to install Sailthru when running the audit script. Similarly, when running this in production, we should not bother installing test requirements. This particular requirement belongs in `sailthru/requirements/test.txt`.

Also, please file a ticket to enable Travis on this repo. There is no point in having tests if we don't run them.

ahsan-ul-haq: @clintonb there is already a ticket to enable Travis on this repo https://openedx.atlassian.net/browse/ECOM-6471

sailthru/sailthru_content/services/catalog_api_service.py
mikedikan: I think you should test it.  In fact, I don't think you need to provide the marketable query string parameter.  If we look at the marketing site code:

https://github.com/edx/edx-mktg/blob/master/docroot/sites/all/modules/features/edx_catalog_api_proxy/edx_catalog_api_proxy.module#L362

We only provide the search query and also pass in the content type (courserun, program) in the query string.  Can you model your approach on this at least and pass in the content type, or even better, find out if the course run search endpoint will give you what you need?

sailthru/sailthru_content/services/catalog_api_service.py
mikedikan: ok, thanks

sailthru/sailthru_content/services/tests/catalog_api_test_mixins.py
mikedikan: Can you explain to me why the dictionary approach is needed at all?  It looks like we do the same for programs, do you have any idea why?

ahsan-ul-haq:
@mikedikan please review.

sailthru/sailthru_content/services/catalog_api_service.py
```diff
@@ -63,25 +63,18 @@ def get_courses(self):
...
-    def get_searchable_course_keys(self):
+    def get_searchable_course_run_keys(self):
         logger.debug('Get Searchable Courses called')
         searchable_course_keys = []
-        searchable_courses = self._get_resource_from_search_api(self.api_client.search.all.facets())
-        for course in searchable_courses:
-            if course.get('key') and course.get('content_type') != 'program':
-                searchable_course_keys.append(course['key'])
+        searchable_course_runs = self._get_resource_from_search_api(
+            self.api_client.search.all.facets(), content_type='courserun'
```
mikedikan: Thank you for changing this to provide the content_type.  

I understand why we make the call to the search endpoint, because we want only the marketable courses.  I don't understand why you make the other call to the courses endpoint to get the course_run data ( courses = get_courses()).  Don't you have all of the information you need from the search endpoint?  I took a quick look at the fields and they all seem to be there.  Can you explain to me why we need to make two separate calls here?

ahsan-ul-haq: We have to make two separate calls as the search endpoint does not have all of the information we need such as seat detail. In search endpoint we can only get the seat type information, not the price, currency and credit provider.

mikedikan: Got it.  After leaving you the comment, I did some digging.  Take a look at the course_run endpoint:
https://prod-edx-discovery.edx.org/api-docs/#!/api/Course_Run_list

You'll see in the swagger documentation that it provides you the 'marketable' flag.  I think you should scrap the dictionary/key approach and just use this endpoint with the marketable flag, is there a reason not to do so?

ahsan-ul-haq: Yes, there is a reason, as you can see https://prod-edx-discovery.edx.org/api/v1/course_runs/?marketable=1 returns 1300 courses whereas on marketing site the number of courses return are 1291 which is same as per count as well as courses on the https://prod-edx-discovery.edx.org/api/v1/search/all/facets/. So we are not sure which 9 marketable courses are not searchable on the marketing site and we are not sure in the future as well. So it seems like https://prod-edx-discovery.edx.org/api/v1/search/all/facets/ is our only reliable endpoint for fetching searchable courses.

mikedikan:
I spoke with @clintonb about this topic and he explained that the difference between the course_runs and search/all/facets endpoints are that the hidden courses will appear in the course_runs endpoint, but not in the search endpoint.  Another option to take here would be to programmatically leave out the hidden course runs in your data to sailthru.

clintonb:
Now that I've thought more about this, please use the Course Runs endpoint. If it doesn't have the filtering we need, we should add it. Using search makes no sense as it simply brings Elasticsearch into play for the purpose of filtering. We should not repeat the mistakes of the current marketing site—using search for filtering/ordering.

ahsan-ul-haq:
@mikedikan please review

ahsan-ul-haq:
@mikedikan I am not sure what is proposing. We cannot only use course_runs endpoint because we need course endpoint as well.

ahsan-ul-haq:
@mikedikan please review
FYI @rock345

mikedikan:
I see now.  You're using the courses endpoint and course_runs endpoint because you need subject/organization data and the course_runs endpoint doesn't have it.  You're using the course_runs endpoint because courses returns you courses/course runs that are not searchable.  I still think we should only use one of these endpoints and add the fields into it that are needed.  I think you could do this by either adding the proper flags into the course endpoint (which I think is only used by sailthru) or by adding the org/subject data into the course_runs endpoint (which has a few other consumers).   Adding the filters to the courses endpoint seems safest because it won't have any performance impact and there is no collision concern, while the course_runs endpoint is being used by other consumers and any new fields should be performance tested.

Please let me know if this is unclear.

ahsan-ul-haq:
@mikedikan I don't think that we could add the course_run endpoint flags in the course endpoint as it depends upon marketing slug and hidden field attributes. It could be possible that one course have multiple runs and some of them are hidden while others are not. So what are we planning to do in that case? Other than that I don't think calling two endpoints should be a concerning thing over here as this is an in-house app which would not have any time concerning effect on the users. 
Apart from that one possible way to reduce the call would be if we add the filters for searchable course_runs here as we have data for both courses and course_runs endpoint.

mikedikan:
@ahsan-ul-haq I think we can likely update the endpoint to return courses only if at least one of the course runs is marketable.  I think that we can do it after.  I'll put in a story to make the fix after you get back from your trip. I'll approve this one, but I don't want that update to slip after.

mikedikan:
Rebase before merging

mikedikan:
https://openedx.atlassian.net/browse/ECOM-6660

PR #38, opened 2016-12-29
Compile updated messages

rlucioni:
When .po files are updated, they need to be compiled to .mo files for use with gettext. This is a project-agnostic approach for accomplishing that.

@edx/ecommerce FYI. This will prevent issues like the one detailed at https://github.com/edx/course-discovery/pull/499#issuecomment-269513662, and has already been run against programs (https://github.com/edx/programs/pull/190).

### DIFF
--- a/transifex/pull.py
+++ b/transifex/pull.py
@@ -88,6 +88,7 @@ class Repo:
         self.owner = None
         self.branch_name = 'update-translations'
         self.message = 'Update translations'
+        self.is_compiled = True
 
     def clone(self):
         """Clone the repo."""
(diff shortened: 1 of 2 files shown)


transifex/pull.py
```diff
@@ -103,8 +104,8 @@ def branch(self):
         """Create and check out a new branch."""
         subprocess.run(['git', 'checkout', '-b', self.branch_name], check=True)
 
-    def pull(self):
-        """Download translated strings from Transifex.
+    def update_translations(self):
```
mikedikan: why the change

rlucioni: Readability and clarity. This is no longer just pulling updated messages from Transifex.

PR #39, opened 2017-01-11
Remove the query string parameters from the marketing url

schenedx:
ECOM-6710

@clintonb @MatthewPiatetsky @umar-nawaz @ahsan-ul-haq Please review

### DIFF
--- a/sailthru/requirements/test.txt
+++ b/sailthru/requirements/test.txt
@@ -3,4 +3,5 @@
 
 ddt==1.0.0
 factory_boy==2.7.0
+pytest==3.0.5
 responses==0.5.1
--- a/sailthru/sailthru_content/services/catalog_api_service.py
+++ b/sailthru/sailthru_content/services/catalog_api_service.py
@@ -46,7 +46,7 @@ class CatalogApiService(object):
 
     def get_courses(self):
         logger.debug('Get Courses called')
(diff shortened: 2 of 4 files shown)


sailthru/sailthru_content/services/sailthru_translation_service.py
```diff
@@ -43,7 +43,7 @@ def _create_course_vars(self, course, course_run, url, site_name, program_dictio
         seat_priority = ['credit', 'professional', 'verified', 'no-id-professional', 'audit', 'honor']
         sailthru_content_vars = {
             'course_run': True,
-            'marketing_url': url,
+            'marketing_url': url.split('?', 1)[0],
```
clintonb: Use the `exclude_utm=1` parameter on the API. See https://prod-edx-discovery.edx.org/api-docs/#!/api/Course_list.

schenedx: Huh, I'll try this out.

schenedx:
@clintonb I took in your feedback, please review again

sailthru/requirements/test.txt
```diff
@@ -4,3 +4,4 @@
 ddt==1.0.0
 factory_boy==2.7.0
 responses==0.5.1
+pytest==3.0.5
```
clintonb: Alphabetize.

PR #40, opened 2017-01-13
Update catalog api service to use new flag marketable_course_runs_only

schenedx:
ECOM-6811

With this PR, I have made the following changes
1. Update the flag used to get the courses and programs from `published_course_runs_only` to use `marketable_course_runs_only`
2. Rename the top level folder from `sailthru` to `edx_sailthru` because the `sailthru` folder name conflicted with the sailthru API client package at https://pypi.python.org/pypi/sailthru-client/2.3.1
3. Updated the searchable course runs endpoint from just the `/course_runs/` to the real search endpoint we expose at `/search/all/facets/`

@edx/ecommerce Please review


rlucioni:
The `marketable_course_runs_only` parameter doesn't currently return "archived" course runs (runs whose end date is in the past, but whose enrollment end date is undefined or in the future). The plan is to change that behavior as part of [ECOM-6819](https://openedx.atlassian.net/browse/ECOM-6819), but if you need the parameter here, that change might have to happen sooner.

schenedx:
@rlucioni Make sense. I believe you and Patrick in Marketing had conversation about this. I also had conversation with Patrick about the use of this new flag. He agree we should use the new flag for courses in Sailthru. I can double check with him about the need for ECOM-6819.

schenedx:
@clintonb OK, I'll create two separate PRs

schenedx:
Closed for https://github.com/edx/ecommerce-scripts/pull/41

rlucioni:
@schenedx ECOM-6819 needs to happen to facilitate use of the enrollment dialog in production. What I'm trying to make clear is that you won't get "archived" courses back if you use the flag right now.

schenedx:
@rlucioni Yes, that's clear. I'll double check with Patrick in marketing again whether he is OK with us not promoting "archived" courses. Last time we talked about this, he expressed that is indeed OK. I'll confirm again anyways.

rlucioni:
@schenedx okay. Jasper wasn't okay with it, which is why we have ECOM-6819.

PR #41, opened 2017-01-13
Rename the top level folder from sailthru to edx_sailthru

schenedx:
Rename the top level folder from sailthru to edx_sailthru because the sailthru folder name conflicted with the sailthru API client package at https://pypi.python.org/pypi/sailthru-client/2.3.1

@edx/ecommerce Please help review


sailthru/sailthru_import/test_sailthru_import_verify.py
```diff
@@ -1,51 +0,0 @@
-import os
```
rlucioni: Why is this being deleted?

schenedx: This is a poor attempt at testing this import script. The whole file is full of bugs. It's not useful so we remove it. It we need this later, we can code a better one

PR #44, opened 2017-01-24
Add marketable flag to the course runs vars ECOM-6892

schenedx:
@edx/ecommerce Please help review
This is to add the `marketable` flag to all the courses updated through this script


edx_sailthru/sailthru_content/services/sailthru_translation_service.py
```diff
@@ -46,7 +46,8 @@ def _create_course_vars(self, course, course_run, url, site_name, program_dictio
             'course_run': True,
             'marketing_url': url,
             'course_id': course['key'],
-            'course_run_id': course_run['key']
+            'course_run_id': course_run['key'],
+            'marketable': 1,  # Mark the course_run to be marketable when updated through this script
```
clintonb: Why `1` instead of `True`?

schenedx: No reason really. I chose `1` because we already have this pattern for variable `activated` on the user object we sync with Sailthru

clintonb: Can we align all of the Boolean values to the same standard—preferably, `True` or `False`? The course_run field above is using a Boolean value. It sounds like marketable and activated are using integers.

schenedx: @clintonb Good point. Let me see what marketing want here, with your point.

rlucioni:
@schenedx what does "marketable" mean in this context? Is it aligned with the way we currently define that filter in the catalog? What do we need to change here if/when that filter changes?

edx_sailthru/sailthru_content/services/sailthru_translation_service.py
schenedx: @clintonb I talked with Jeanie and she agreed we should stick with "0" and "1" values. I'll make the update to "course_run" variable as well to use "0" and "1" value.

schenedx:
@rlucioni To answer directly, the "marketable" variable tells our email marketing users whether this course in the right state to be advertised to our learners. We have this bit in catalog, and we want this in Sailthru to match what we exposes in catalog.
However, your question on scenario when filter changes, is a good one. Upon further consideration, I think we should expand the scope of this change a little to keep the changes in sync. Let me talk to Jeanie again.

rlucioni:
@schenedx the reason I'm asking is because you can't currently determine whether a course run should be advertised by using only the catalog's `marketable` filter. There are edge cases concerning course and enrollment dates that aren't covered.

schenedx:
@rlucioni That's fair, I think in this case, "marketable" means the course is returned from the "marketable_course_runs_only" flag and searchable (Returns from the search api just like the marketing site). We don't worry much about edge cases because those are not that important to email marketing at the moment.

schenedx:
@rlucioni The above understanding is not reflected on this PR, I'll update the logic in the sailthru sync script to make this clearer

rlucioni:
Okay. That aside, I'm not sure why this change is necessary. If the only course runs returned by the catalog are those deemed `marketable`, why is a new field necessary? Shouldn't the presence of the run be enough? If not, then when is the value of this field set to 0/False?

schenedx:
@edx/ecommerce I updated the PR with the most recent understanding. Please help me review.

With this change, we are going to update all the course runs in Sailthru. During this process, we will be updating marketable courses with the flag `marketable` and value `1`. This would satisfy the requirement that we want all of our courses in Sailthru at the same time a way to get courses we want to advertise and promote thru email marketing.

A big assumption that I believe I was wrong on, was that with Sailthru sync script, we only updated and added new courses to the content library. We never deleted courses from Sailthru content library. With this change, we can now make sure we only advertise those ones we want to advertise.

PR #46, opened 2017-02-08
ECOM-6899 Populate reporting_type and announcement fields on course runs

MatthewPiatetsky:
@edx/ecommerce


populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,80 @@
+import argparse
```
MatthewPiatetsky: the root directory has a hyphen, I think for that reason I wasn't able to do a relative or absolute import, is there any way to do that other than changing the structure of the whole repo?  
I don't know all the different jenkins jobs running these scripts and changing the root directory would break all of them

clintonb: Why does the rest of the repo need to change? Just move these files into their own module.

MatthewPiatetsky: this file is using the service file which is also used by other scripts. If I move this into the course_discovery folder I'm not sure how I can import that service (other than maybe using __import__ but this seems less hacky than that)

clintonb: Try a relative import: `from ..services`. Otherwise, we need to bite the bullet and fix this repo.

MatthewPiatetsky: I get this error if I try to do a relative import  
SystemError: Parent module '' not loaded, cannot perform relative import
if you're not sure how to fix this that's fine and we can refactor the repo, but I would like to do that with a new groomed/planned story that we groom/plan with Simon to get input on the jenkins jobs. If you think we should block this PR on that we can mark the subtask as blocked. Thoughts?

clintonb: I would prefer we do this correctly now to avoid kicking the can down the road. Please work with @schenedx to make that happen.

MatthewPiatetsky: a lot more than just sailthru uses this repo. I'll work with the team

clintonb: The only other automated task is the Transifex task built by @rlucioni. The rest are one-off scripts.

populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,80 @@
...
+import datetime
+from dateutil.parser import parse
+import json
+import logging
+import os
+
+import logging
+
+from edx_sailthru.sailthru_content.services.catalog_api_service import CatalogApiService
```
clintonb: Another alternative here is to simply not share code with this module. Just use `EdXRestApiClient` directly. There is nothing special about what the Sailthru code is doing.

MatthewPiatetsky: true, do you have a preference around doing this or doing the refactor?

clintonb: Just make your own client. We can refactor later.

MatthewPiatetsky: Ok I've moved the client to the file

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,114 @@
...
+            self.access_token, expires = EdxRestApiClient.get_oauth_access_token(
+                '{root}/access_token'.format(root=oauth_host),
+                oauth_key,
+                oauth_secret, token_type='jwt'
+            )
+        except Exception:
+            logger.exception('No access token acquired through client_credential flow.')
+            raise
+
+        logger.info('Token retrieved: %s', access_token)
```
clintonb: NEVER log the token. That's equivalent to logging a password in plaintext. Remove this from here and where it was copied from.

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,114 @@
...
+
+        self.api_client = EdxRestApiClient(api_url_root, jwt=self.access_token)
+
+    def _get_resource_from_api(self, api_endpoint, page_size=None, **kwargs):
```
clintonb: This is not used. Remove it. It can be restored if it's actually needed in the future.

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,114 @@
...
+
+import logging
+
+from edx_rest_api_client.client import EdxRestApiClient
+
+
+logger = logging.getLogger()
+
+
+class CatalogApiService(object):
```
clintonb: Move this to its own module, `services.py`.

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,114 @@
...
+
+        return results
+
+def process_line(line):
+    key, reporting_type, announcement = line.strip().split(',')
+    data = {
+        'reporting_type': reporting_type
+    }
+    if announcement != '':
+        # Convert 2/1/2016 to 2016-02-01T00:00:00+00:00
```
clintonb: The conversion is implied by the calls to `parse()` and `isoformat()`.

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,114 @@
...
+            results.extend(response.get('results'))
+
+        return results
+
+def process_line(line):
+    key, reporting_type, announcement = line.strip().split(',')
+    data = {
+        'reporting_type': reporting_type
+    }
+    if announcement != '':
```
clintonb: `if announcement:`

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,114 @@
...
+    try:
+        response = course_runs_endpoint.patch(data=data)
+        logging.info("{} update succeeded".format(key))
+    except Exception as e:
+        logging.error("Catalog returned this error: {}".format(e.response.content))
+        logging.error("{} update failed".format(key))
+
+def main():
+    parser = argparse.ArgumentParser(description='Populates reporting type and announcment on course runs')
+    parser.add_argument('--filename', help='Name of file with data on course run tags and announcment dates.')
```
clintonb: s/tags/types/

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,114 @@
...
+        '--catalog_api_url',
+        default=os.environ.get('CATALOG_API_URL', None),
+        help='Content api url.'
+    )
+
+
+    args = parser.parse_args()
+
+    catalog_api_service = CatalogApiService(
+        None,
```
clintonb: Remove.

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,114 @@
...
+        args.oauth_host,
+        args.oauth_key,
+        args.oauth_secret,
+        args.catalog_api_url
+    )
+
+    with open(args.filename) as f:
+        next(f)
```
clintonb: Use [`csv.DictReader`](https://docs.python.org/3/library/csv.html#csv.DictReader).

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+    parser.add_argument(
+        '--oauth_secret',
+        default=os.environ.get('CATALOG_OAUTH_SECRET', None),
+        help='OAuth2 secret token used to authenticate Course Content API calls.'
+    )
+    parser.add_argument(
+        '--catalog_api_url',
+        default=os.environ.get('CATALOG_API_URL', None),
+        help='Content api url.'
```
clintonb: Catalog API URL

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,114 @@
...
+    parser.add_argument('--filename', help='Name of file with data on course run tags and announcment dates.')
+    parser.add_argument(
+        '--oauth_host',
+        default=os.environ.get('CATALOG_OAUTH_HOST', 'https://api.edx.org/oauth2/v1'),
+        help='OAuth2 base url.'
+    )
+    parser.add_argument(
+        '--oauth_key',
+        default=os.environ.get('CATALOG_OAUTH_KEY', None),
+        help='OAuth2 key used to authenticate Course Content API calls.'
```
clintonb: OAuth client key

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+    parser.add_argument(
+        '--oauth_secret',
+        default=os.environ.get('CATALOG_OAUTH_SECRET', None),
+        help='OAuth2 secret token used to authenticate Course Content API calls.'
```
clintonb: OAuth client secret

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,114 @@
...
+        logging.error("Catalog returned this error: {}".format(e.response.content))
+        logging.error("{} update failed".format(key))
+
+def main():
+    parser = argparse.ArgumentParser(description='Populates reporting type and announcment on course runs')
+    parser.add_argument('--filename', help='Name of file with data on course run tags and announcment dates.')
+    parser.add_argument(
+        '--oauth_host',
+        default=os.environ.get('CATALOG_OAUTH_HOST', 'https://api.edx.org/oauth2/v1'),
+        help='OAuth2 base url.'
```
clintonb: OAuth access token endpoint URL

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,114 @@
...
+        logging.info("{} update succeeded".format(key))
+    except Exception as e:
+        logging.error("Catalog returned this error: {}".format(e.response.content))
+        logging.error("{} update failed".format(key))
+
+def main():
+    parser = argparse.ArgumentParser(description='Populates reporting type and announcment on course runs')
+    parser.add_argument('--filename', help='Name of file with data on course run tags and announcment dates.')
+    parser.add_argument(
+        '--oauth_host',
```
clintonb: oauth_access_token_url

We only use this value to get the access token. Just get the actual access token endpoint instead of piecing it together.

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,114 @@
...
+        default=os.environ.get('CATALOG_OAUTH_HOST', 'https://api.edx.org/oauth2/v1'),
```
clintonb: `default=os.environ.get('CATALOG_OAUTH_ACCESS_TOKEN_URL', 'https://api.edx.org/oauth2/v1/access_token'),`

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,114 @@
...
+        data['announcement'] = announcement
+    return key, data
+
+def update_course_run(key, data, catalog_api_service):
+    course_runs_endpoint = catalog_api_service.api_client.course_runs(key)
+    try:
+        response = course_runs_endpoint.patch(data=data)
+        logging.info("{} update succeeded".format(key))
+    except Exception as e:
+        logging.error("Catalog returned this error: {}".format(e.response.content))
```
clintonb: `logging.exception('Failed to update course run [%s]', key)`

If you're catching all `Exception` classes, you cannot assume the raised exception has a `response` attribute.

MatthewPiatetsky:
@clintonb is this change what you were looking for/did I address your comments?

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,114 @@
+import argparse
+import datetime
+from dateutil.parser import parse
+import json
+import logging
+import os
+
+import logging
```
clintonb: Remove.

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,114 @@
+import argparse
+import datetime
+from dateutil.parser import parse
```
clintonb: I'm fairly certain this is a third-party import, and should be grouped appropriately.

schenedx:
@MatthewPiatetsky Why do we do this operation as a python script rather than a management command on Course-Discovery IDA?

course_discovery/populate_course_run_reporting_type_and_announcement.py
MatthewPiatetsky: Personally, I wasn't exactly sure what formatting these did until I wrote this, to me this would be helpful if I don't already know what isoformat is which i didn't

MatthewPiatetsky:
@schenedx because this is a one time script that isn't specific to a need we have in the catalog. We added a patch endpoint to the rest api, so any random tasks that aren't generic enough to belong in course_discovery don't have to go there

MatthewPiatetsky:
@clintonb i think i addressed your latest comments

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,74 @@
...
+    catalog_api_service = CatalogApiService(
+        args.oauth_access_token_url,
+        args.oauth_key,
+        args.oauth_secret,
+        args.catalog_api_url
+    )
+
+    with open(args.filename) as f:
+        reader = csv.DictReader(f)
+        next(reader, None)
```
clintonb: This is not necessary.

course_discovery/services.py
```diff
@@ -0,0 +1,23 @@
+import logging
+
+from edx_rest_api_client.client import EdxRestApiClient
+
+
+logger = logging.getLogger()
+
+
+class CatalogApiService(object):
+    """The service to interface with edX catalog API"""
```
clintonb: Catalog

edx_sailthru/sailthru_content/services/catalog_api_service.py
```diff
@@ -25,8 +25,6 @@ def __init__(self, access_token, oauth_host, oauth_key, oauth_secret, api_url_ro
                 logger.exception('No access token provided or acquired through client_credential flow.')
                 raise
 
-            logger.info('Token retrieved: %s', access_token)
```
clintonb: `logger.info('Retrieved access token.')`

course_discovery/services.py
```diff
@@ -0,0 +1,25 @@
+import logging
+
+from edx_rest_api_client.client import EdxRestApiClient
+
+
+logger = logging.getLogger()
+
+
+class CatalogApiService(object):
```
schenedx: @MatthewPiatetsky We already have a `CatalogApiService` at https://github.com/edx/ecommerce-scripts/blob/master/edx_sailthru/sailthru_content/services/catalog_api_service.py
I know what you want here is super simple. I would argue the one in `edx_sailthru` can use this service here. Is there a good reason or consideration to have these CatalogApiService independent?

MatthewPiatetsky: I was originally using the existing service and had my script at the top level of the repo. I couldn't have it in a sub directory because the top folder in the repo has a hyphen which breaks absolute python imports. 
Clinton suggested refactoring the whole repo, but since that would involve updating a bunch of jobs, he suggested alternatively to just make a new client

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,73 @@
...
+        # Convert 2/1/2016 to 2016-02-01T00:00:00+00:00
+        announcement = parse(announcement).isoformat()
+        data['announcement'] = announcement
+    return line['course_run_key'], data
+
+def update_course_run(key, data, catalog_api_service):
+    course_runs_endpoint = catalog_api_service.api_client.course_runs(key)
+    try:
+        response = course_runs_endpoint.patch(data=data)
+        logging.info("{} update succeeded".format(key))
```
clintonb: Nit: single quotes

course_discovery/populate_course_run_reporting_type_and_announcement.py
```diff
@@ -0,0 +1,73 @@
...
+        help='OAuth client key.'
+    )
+    parser.add_argument(
+        '--oauth_secret',
+        default=os.environ.get('CATALOG_OAUTH_SECRET', None),
+        help='OAuth2 secret token.'
+    )
+    parser.add_argument(
+        '--catalog_api_url',
+        default=os.environ.get('CATALOG_API_URL', None),
```
clintonb: Why not assume production, as you do with the access token URL?

course_discovery/services.py
schenedx: Got it. Can you create a dev task story and label it with `quick-win` so it gets tracked by the Spartans team? Thank you

PR #47, opened 2017-03-13
Update the sailthru synch script to use a smaller batch size for catalog api query

schenedx:
@rlucioni @MatthewPiatetsky Please help review

Just decrease the catalog query batch size from 500 to 100.
This certainly makes the Sailthru synch script run last longer, but that's intended


schenedx:
@rlucioni Will do. Thanks

PR #49, opened 2017-04-27
WIP: Run Tests in ecommerce-scripts on Travis

adeelkhan:
ECOM-6471

Adds Travis CI support for ecommerce-script repo.
Not for external review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: python
+python:
+    - "2.7"
+    - "3.5"
+    - "3.6"
+
+install:
+  - pip install -r ./edx_sailthru/requirements/test.txt
+
+script:
+  - pytest
--- a/edx_sailthru/requirements/test.txt
+++ b/edx_sailthru/requirements/test.txt
@@ -5,3 +5,6 @@ ddt==1.0.0
 factory_boy==2.7.0
 pytest==3.0.5
 responses==0.5.1
+Faker==0.7.3
+mock==2.0.0
+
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -0,0 +1,9 @@
+language: python
+python:
+    - "2.7"
```
clintonb: Test for both Python 2.7, 3.5, and 3.6.

.travis.yml
```diff
@@ -0,0 +1,9 @@
...
+
+install:
+  - pip install -r ./edx_sailthru/requirements/test.txt
+
+script:
+  - pytest
```
clintonb: Does this work from the root of the repo?

.travis.yml
adeelkhan: @clintonb Tested it for 2.7.10. Its mentioned in README.rst that its designed for 2.7.x

clintonb: Python 2.7 is about to die. All of our code needs to work for Python 3.5+.

.travis.yml
adeelkhan: yes

.travis.yml
clintonb: Clarification: you don't need to make all of the code Python 3.x-compatible with this PR, but we should still be testing to ensure compatibility.

adeelkhan: @clintonb Ok. Made changes and now works for 3.6 too.

edx_sailthru/requirements/test.txt
```diff
@@ -5,3 +5,5 @@ ddt==1.0.0
 factory_boy==2.7.0
 pytest==3.0.5
 responses==0.5.1
+Faker==0.7.3
+mock==2.0.0
```
clintonb: Add a newline

PR #50, opened 2017-05-16
Add script for identifying verified seats without upgrade deadlines

rlucioni:
Walks through all course runs known to the discovery service and finds the ones with verified seats which are missing upgrade deadlines.

LEARNER-754

@edx/learner-growth FYI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,10 @@ target/
 
 .idea/
 
+# pyenv
+.python-version
+
+# macOS
 .DS_Store
 
 # Google Analytics Specific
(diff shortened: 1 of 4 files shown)


rlucioni:
@clintonb there's some overhead involved in doing that that I'd like to avoid right now. Could we see how this develops and consider making it a management command if there's a clear need? As things stand, this is looking like it'll be a one-off that's really only useful when you're operating a system with thousands of course runs. I'm not sure if anyone besides us needs this right now.

(Also, although at the moment all the script does is find course runs whose seats need to be updated, I'm planning to build on it by adding code which manipulates the CAT to actually perform the updates. Having a management command do that strikes me as a odd. We could add a PATCH endpoint to ecommerce and use that, but again, there's overhead there we can do without right now.)

PR #51, opened 2017-05-18
Improve pagination on the script, and make it python 3.5 compatible

schenedx:
I wanted to improve the pagination logic on the script.
I also want to make sure it is python 3.5 compatible.

Please review @clintonb @ahsan-ul-haq @adeelkhan

### DIFF
--- a/edx_sailthru/sailthru_content/services/catalog_api_service.py
+++ b/edx_sailthru/sailthru_content/services/catalog_api_service.py
@@ -1,7 +1,7 @@
 import logging
+from urllib.parse import urlparse, parse_qs
 
 from edx_rest_api_client.client import EdxRestApiClient
-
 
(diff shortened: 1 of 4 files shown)


edx_sailthru/sailthru_content/services/catalog_api_service.py
```diff
@@ -30,38 +30,35 @@ def __init__(self, access_token, oauth_host, oauth_key, oauth_secret, api_url_ro
...
+            else:
+                break
 
         return results
 
     def get_courses(self):
         logger.debug('Get Courses called')
         return self._get_resource_from_api(
             self.api_client.courses(),
-            page_size=COURSES_PAGE_SIZE,
```
clintonb: Going forward, we want to use page number pagination, not limit-offset.

PR #52, opened 2017-05-18
Use the CAT to correct upgrade deadlines

rlucioni:
LEARNER-754


rlucioni:
@mikedikan I do. I've already run a few dry runs.

PR #53, opened 2017-06-01
Fix the loading of the fixup file

schenedx:
With python 3.5, we needed to update the way we load the fix_up file.
@rlucioni @clintonb Please help review

### DIFF
--- a/edx_sailthru/sailthru_content/sync_library.py
+++ b/edx_sailthru/sailthru_content/sync_library.py
@@ -96,7 +96,7 @@ def load_fixups(filename):
     if not filename:
         return []
 
-    with open(filename, 'rb') as f:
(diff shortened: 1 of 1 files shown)


schenedx:
I am merging to unblock myself.

edx_sailthru/sailthru_content/sync_library.py
```diff
@@ -96,7 +96,7 @@ def load_fixups(filename):
     if not filename:
         return []
 
-    with open(filename, 'rb') as f:
+    with open(filename, 'rt') as f:
```
rlucioni: @schenedx you shouldn't need either of these if you're just trying to read the file. `open(filename) as f:` would be enough.

PR #54, opened 2017-06-09
Add ability to skip message compilation to transifex/pull.py

amangano-edx:
Adds a new option, `--skip-compile-messages`, that when passed to `transifex/pull.py` will result in the script skipping the compilemessages step. This makes it possible to use this script with the marketing site repository, which does not need to compile .po files.

### DIFF
--- a/transifex/pull.py
+++ b/transifex/pull.py
@@ -14,22 +14,27 @@ settings.yaml, simply run:
 If you want to run the script for a specific repo, provide a clone URL as an argument:
 
     python transifex/pull.py git@github.com:edx/course-discovery.git
+
+If you want to skip the compile messages step, pass the --skip-compilemessages option.
+
+    python transifex/pull.py git@github.com:edx/course-discovery.git --skip-compilemessages
 """
-import concurrent.futures
-from contextlib import contextmanager
 import logging
-from logging.config import dictConfig
 import os
-from os.path import join, abspath, dirname
 import re
 import subprocess
 import sys
 import time
+from argparse import ArgumentParser
+from contextlib import contextmanager
+from logging.config import dictConfig
+from os.path import abspath, dirname, join
 from urllib.parse import urlparse
 
-from github import Github, GithubException
 import yaml
 
+import concurrent.futures
+from github import Github, GithubException
 
 # Configure logging.
 dictConfig({
@@ -71,7 +76,7 @@ edx = github.get_organization('edx')
 
 class Repo:
     """Utility representing a Git repo."""
-    def __init__(self, clone_url):
+    def __init__(self, clone_url, skip_compilemessages=False):
         # See https://github.com/blog/1270-easier-builds-and-deployments-using-git-over-https-and-oauth.
         parsed = urlparse(clone_url)
         self.clone_url = '{scheme}://{token}@{netloc}{path}'.format(
@@ -88,7 +93,8 @@ class Repo:
         self.owner = None
         self.branch_name = 'update-translations'
         self.message = 'Update translations'
-        self.is_compiled = True
+        self.skip_compilemessages = skip_compilemessages
+        self.compilemessages_failed = False
 
     def clone(self):
         """Clone the repo."""
@@ -115,6 +121,10 @@ class Repo:
         """
         subprocess.run(['make', 'pull_translations'], check=True)
 
+        if self.skip_compilemessages:
+            logger.info('Skipping compilemessages.')
+            return
+
         # Messages may fail to compile (e.g., a translator may accidentally translate a
         # variable in a Python format string). If this happens, we want to proceed with
         # the PR process and notify the team that messages failed to compile.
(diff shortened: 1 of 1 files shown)


transifex/pull.py
```diff
@@ -115,21 +120,22 @@ def update_translations(self):
...
-            # doesn't require DJANGO_SETTINGS_MODULE to be defined when run from the project root,
-            # and also doesn't need django.setup() to be run. Because of this, we can get away with
-            # django-admin.py instead of manage.py. The latter defines a default value for
-            # DJANGO_SETTINGS_MODULE and causes django.setup() to run, which is undesirable here for reasons
-            # ranging from Python 2/3 incompatibility errors across projects to forcing the installation
-            # of packages which provide installed apps custom to each project.
-            subprocess.run(['django-admin.py', 'compilemessages'], check=True)
-        except subprocess.CalledProcessError:
-            self.is_compiled = False
+        if not skip_compile_messages:
```
clintonb: Instead of indenting everything, just return early:

```python
if skip_compile_messages:
  # TODO Log the fact that compilation is skipped
  return
```

transifex/pull.py
```diff
@@ -14,7 +14,12 @@
 If you want to run the script for a specific repo, provide a clone URL as an argument:
 
     python transifex/pull.py git@github.com:edx/course-discovery.git
+
+If you want to skip the compile messages step, pass the --skip-compile-messages option.
+
+    python transifex/pull.py git@github.com:edx/course-discovery.git --skip-compile-messages
 """
+from argparse import ArgumentParser
```
clintonb: Order your imports properly. You might want to run `isort` on this file.

transifex/pull.py
```diff
@@ -223,7 +229,7 @@ def pull(repo):
                 logger.info('No changes detected for [%s]. Cleaning up.', repo.name)
 
         if pr:
-            if not repo.is_compiled:
+            if (not skip_compile_messages) and (not repo.is_compiled):
```
rlucioni: Minor: for clarity, consider using the existing `self.is_compiled` a little differently. You could just default it to False, for example, only setting it True once compilation completes, if it's run at all.

transifex/pull.py
```diff
@@ -288,12 +294,28 @@ def pull(repo):
         repo.cleanup(pr)
 
 
+def parse_arguments():
+    parser = ArgumentParser()
+    parser.add_argument(
+        'clone_url',
+        nargs='?',
+        help='Url to use to clone the repository. If blank, urls will be pulled from settings.yaml'
```
rlucioni: Minor: URL, URLs

transifex/pull.py
amangano-edx: Changed `is_compiled` to `compilemessages_failed`, since it is only used to flag when message compilation fails. That makes this if statement a lot clearer.

transifex/pull.py
amangano-edx: Changed.

transifex/pull.py
amangano-edx: Changed.

transifex/pull.py
amangano-edx: Fixed.

amangano-edx:
@rlucioni @clintonb I think this should be ready for a second look. How have we been testing these scripts before shipping them to prod?

clintonb:
You can test this locally, by installing the package from Git in your virtualenv, and using the commands as normal.

transifex/pull.py
```diff
@@ -288,12 +297,28 @@ def pull(repo):
...
-    try:
-        clone_url = sys.argv[1]
-        repo = Repo(clone_url)
-        pull(repo)
-    except IndexError:
+    args = parse_arguments()
+
+    if args.clone_url:
+        repo = Repo(args.clone_url)
+        pull(repo, skip_compile_messages=args.skip_compile_messages)
```
rlucioni: @amangano-edx small suggestion: instead of threading this variable through as a kwarg, you could just pass it when initializing the `Repo` above.

amangano-edx: I think I'd prefer to leave it as is, as I suspect I will be refactoring a lot of the Repo code when I write the push script, as much of it will be shared between the two files.

amangano-edx: Actually, I decided to make this change after all.

transifex/pull.py
```diff
@@ -302,5 +328,5 @@ def pull(repo):
 
         with concurrent.futures.ProcessPoolExecutor() as executor:
             for clone_url in settings['repos']:
-                repo = Repo(clone_url)
+                repo = Repo(clone_url, args.skip_compilemessages)
```
rlucioni: Minor: this is a keyword argument and should be used as such! Since this is Python 3, you can force this by writing the `__init__` method as:
```py
def __init__(self, clone_url, *, skip_compilemessages=False):
```

amangano-edx: Fixed.

PR #55, opened 2017-06-16
Refactor transifex pull

amangano-edx:
Refactors transifex/pull.py to extract functionality that will be useful in writing other transifex/translation scripts, such as transifex/push.py, which will be used to extract translations and push them to transifex.

JIRA: https://openedx.atlassian.net/browse/LEARNER-1350

### DIFF
--- a/transifex/pull.py
+++ b/transifex/pull.py
@@ -6,12 +6,7 @@ commit, push, and merge them to their respective repos.
 To use, export an environment variable `GITHUB_ACCESS_TOKEN`. The token requires
 GitHub's "repo" scope.
 
-Run the script from the root of this repo. To use the repo clone URLs stored in
-settings.yaml, simply run:
-
-    python transifex/pull.py
-
-If you want to run the script for a specific repo, provide a clone URL as an argument:
+Run the script from the root of this repo.
 
     python transifex/pull.py git@github.com:edx/course-discovery.git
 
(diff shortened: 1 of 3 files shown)


transifex/pull.py
```diff
@@ -317,8 +83,7 @@ def parse_arguments():
     args = parse_arguments()
 
     if args.clone_url:
-        repo = Repo(args.clone_url, skip_compilemessages=args.skip_compilemessages)
-        pull(repo)
+        pull(args.clone_url, skip_compilemessages=args.skip_compilemessages)
     else:
         logger.info('No arguments provided. Using settings.yaml.')
```
amangano-edx: @rlucioni do we still need to support this? It doesn't seem to work when I run the script like this locally from master. I'd prefer to remove this section if it is not being used by anyone as it makes it more complicated to add options like `--merge-method`, as you would probably want a different one for each repo.

rlucioni:
@amangano-edx can you delete `transifex/utils/__init__.py`? Python 3 doesn't require these `__init__.py` files. For more, see [PEP 420](https://www.python.org/dev/peps/pep-0420/).

transifex/pull.py
rlucioni: @amangano-edx go ahead and remove it. When the script was initially developed it was run like this, but now we use Jenkins jobs to invoke the script for a specific repo.

transifex/pull.py
```diff
@@ -19,225 +19,40 @@
...
-from contextlib import contextmanager
-from logging.config import dictConfig
 from os.path import abspath, dirname, join
-from urllib.parse import urlparse
 
 import yaml
 
 import concurrent.futures
-from github import Github, GithubException
+from utils.common import logger, repo_context
```
rlucioni: Given that the script is run with:
```
$ python transifex/pull.py <URL>
```
I think you may want to import `from transifex.utils.common ...`. You may need that to get rid of `__init__.py`.

transifex/pull.py
```diff
@@ -19,225 +19,40 @@
...
-import time
 from argparse import ArgumentParser
-from contextlib import contextmanager
-from logging.config import dictConfig
 from os.path import abspath, dirname, join
-from urllib.parse import urlparse
 
 import yaml
 
 import concurrent.futures
```
rlucioni: This import can be deleted, too.

transifex/utils/common.py
```diff
@@ -0,0 +1,283 @@
+import logging
```
rlucioni: There are no "non-common" utils. Simplify a bit and put this in `transifex/utils.py`.

transifex/utils/common.py
```diff
@@ -0,0 +1,283 @@
...
+                    logger.info(
+                        'Status checks on [%s/#%d] are pending. This is retry [%d] of [%d].',
+                        self.name, self.pr.number, retries, self.MAX_MERGE_RETRIES
+                    )
+
+                    retries += 1
+
+                    if retries <= self.MAX_MERGE_RETRIES:
+                        # Exponential backoff.
+                        time.sleep(2 ** retries)
```
rlucioni: Given that you're in here refactoring, consider changing this to poll every 5 or 10 minutes. I've noticed that the 10 retries we currently make (34 minutes of sleep time) are *just* short of the time required to merge some PRs. If we bump the retry count to 11 we'll end up waiting `(2^(11+1) - 1) / 60 = 68.25` minutes, which is way more than necessary.

transifex/utils/common.py
```diff
@@ -0,0 +1,283 @@
...
+       everything up.
+    """
+    repo = Repo(*args, **kwargs)
+    try:
+        repo.clone()
+        with cd(repo):
+            repo.branch()
+            yield repo
+    finally:
+        repo.cleanup()
```
rlucioni: I would add a note about this being a good idea because it cleans up the (Jenkins) workspace even in the event of a failure, allowing the script to be rerun without issue.

amangano-edx:
@rlucioni I think I've addressed your feedback. Please look again when you have a moment.

transifex/utils.py
```diff
@@ -0,0 +1,285 @@
...
+# see http://pygithub.github.io/PyGithub/v1/reference.html.
+github_access_token = os.environ['GITHUB_ACCESS_TOKEN']
+github = Github(github_access_token)
+edx = github.get_organization('edx')
+
+
+class Repo:
+    # Combined with exponential backoff, limiting merge retries to 10 results in
+    # a total 34 minutes of sleep time. Status checks should almost always
+    # complete in this period.
```
rlucioni: Please update this comment to reflect the new 5-minute polling strategy.

transifex/pull.py
```diff
@@ -6,238 +6,44 @@
...
-                    '-c', 'user.email={}'.format(os.environ['GIT_USER_EMAIL']),
-                    'commit', '-m', self.message
-                ],
-                check=True
-            )
 
-    def push(self):
-        """Push branch to the remote."""
-        subprocess.run(['git', 'push', '-u', 'origin', self.branch_name], check=True)
+from utils import logger, repo_context
```
rlucioni: It looks like Python is smart enough not to require this, but I tend to write fully qualify these import paths for clarity. In this case, that would be `from transifex.utils import ...`.

transifex/pull.py
```diff
@@ -6,238 +6,44 @@
...
-def pull(repo):
+def pull(clone_url, skip_compilemessages=False):
     """Pulls translations for the given repo.
 
     If applicable, commits them, pushes them to GitHub, opens a PR, waits for
     status checks to pass, then merges the PR and deletes the branch.
     """
-    pr = None
-    logger.info('Pulling translations for [%s].', repo.name)
+    with repo_context(clone_url) as repo:
```
BenjiLee: @amangano-edx Is this is the only code you added?

transifex/pull.py
amangano-edx: I tried doing that but got an ImportError when running locally.

transifex/pull.py
amangano-edx: No, I also moved lots of code around, changed the sleep time for merge retries, removed the ability to run the script without an explicit clone_url, among other things.

transifex/utils.py
amangano-edx: Updated.

PR #56, opened 2017-06-16
Add support for custom merge methods to translation scripts

amangano-edx:
These changes enable us to use a custom merge method when merging PRs produced via the translation scripts. This is necessary since several of our repositories (credentials, course-discovery, ecommerce, edx-mktg) do not allow merge commits.

Depends on https://github.com/edx/ecommerce-scripts/pull/55

JIRA: https://openedx.atlassian.net/browse/LEARNER-1089 

Note: it should only be necessary to review https://github.com/edx/ecommerce-scripts/pull/56/commits/1f805f88857823d2ec23c298205a918125384ecd as all the other commits are from https://github.com/edx/ecommerce-scripts/pull/55

### DIFF
--- a/transifex/pull.py
+++ b/transifex/pull.py
@@ -19,20 +19,13 @@ If you want to skip the compile messages step, pass the --skip-compilemessages o
 
     python transifex/pull.py git@github.com:edx/course-discovery.git --skip-compilemessages
 """
-import time
 from argparse import ArgumentParser
 from os.path import abspath, dirname, join
 
 import yaml
-from github import GithubException
 
 import concurrent.futures
 from utils.common import Repo, cd, logger
-
-# Combined with exponential backoff, limiting retries to 10 results in
-# a total 34 minutes of sleep time. Status checks should almost always
-# complete in this period.
-MAX_RETRIES = 10
 
 
 def pull(repo):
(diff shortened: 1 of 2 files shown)


transifex/utils/common.py
```diff
@@ -67,11 +67,45 @@ def repo_context(*args, **kwargs):
         repo.cleanup()
 
 
+# Merge methods supported by the Github API. https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button
+MERGE_METHODS = {'merge', 'squash', 'rebase'}
+
+
+GithubPullRequest = github.PullRequest.PullRequest
+class ExtendedPullRequest(GithubPullRequest):
+    def merge(self, commit_message=github.GithubObject.NotSet, merge_method=None):
```
rlucioni: I would default `merge_method` to `rebase`. As you noted, most of our repos require this, and in those that don't, rebase merging is still allowed (and preferable to creating merge bubbles).

transifex/pull.py
```diff
@@ -93,4 +102,4 @@ def parse_arguments():
 
         with concurrent.futures.ProcessPoolExecutor() as executor:
```
rlucioni: Like I noted in https://github.com/edx/ecommerce-scripts/pull/55, you can just get rid of this block.

amangano-edx:
Replaced by https://github.com/edx/ecommerce-scripts/pull/60

PR #57, opened 2017-06-16
Adds transifex push script

amangano-edx:
Adds script to extract translations strings and push them to transifex.

WIP because I have not tested this yet.

Depends on https://github.com/edx/ecommerce-scripts/pull/56 and https://github.com/edx/ecommerce-scripts/pull/55

Note: it should only be necessary to review https://github.com/edx/ecommerce-scripts/commit/d3688a9155f67f59b7963dcd4b58ee32e1c0959b, as all the other commits are from https://github.com/edx/ecommerce-scripts/pull/56 and https://github.com/edx/ecommerce-scripts/pull/55

### DIFF
--- a/transifex/push.py
+++ b/transifex/push.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python3
+"""
+This script can be used to automatically extract translation strings,
+commit, push, and merge them to their respective repos, and then push them to transifex.
+
+To use, export an environment variable `GITHUB_ACCESS_TOKEN`. The token requires
+GitHub's "repo" scope.
+
+Run the script from the root of this repo.
+
+    python3 transifex/push.py git@github.com:edx/course-discovery.git
+
+If you want to use a custom merge method pass the --merge-method option.
+
+    python3 transifex/push.py git@github.com:edx/course-discovery.git --merge-method rebase
+"""
+from argparse import ArgumentParser
+from utils.common import MERGE_METHODS, logger, repo_context
+
+
+def push(clone_url, merge_method=None):
+    """Extracts translations for the given repo, commits them, pushes them to GitHub, opens a PR, waits for status
+        checks to pass, merges the PR, deletes the branch, and pushes the updated translation files to transifex.
+    """
+    with repo_context(clone_url, merge_method=merge_method) as repo:
+        logger.info('Extracting translations for [%s].', repo.name)
+        repo.extract_translations()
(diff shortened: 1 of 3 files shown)


transifex/push.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python3
+"""
+This script can be used to automatically extract translation strings,
+commit, push, and merge them to their respective repos, and then push them to transifex.
```
rlucioni: s/transifex/Transifex, here and throughout.

transifex/push.py
```diff
@@ -0,0 +1,58 @@
...
+
+To use, export an environment variable `GITHUB_ACCESS_TOKEN`. The token requires
+GitHub's "repo" scope.
+
+Run the script from the root of this repo.
+
+    python3 transifex/push.py git@github.com:edx/course-discovery.git
```
rlucioni: s/python3/python

transifex/push.py
```diff
@@ -0,0 +1,58 @@
...
+
+# The name of the branch to use.
+BRANCH_NAME = 'update-translation-strings'
+
+# The commit message to use.
+MESSAGE = 'Update translation strings'
+
+
+def push(clone_url, merge_method=None):
+    """Extracts translations for the given repo, commits them, pushes them to GitHub, opens a PR, waits for status
```
rlucioni: I'm unclear on why it's necessary for this script to do anything other than push translation files to Transifex. Could you explain to me again why merges of commits to edx-mktg with unextracted translations being allowed?

amangano-edx: Since the marketing site repository is private, we cannot easily/reliably setup automatic source translation updates on Transifex. Instead, we plan to use this job to manually push updates to Transifex on a regular schedule. I do not see any benefit to requiring devs to extract_translations before every merge, as the updated strings would not make it to Transifex until this job is run anyway. Given that, I don't think it makes sense to add an extra step to the marketing site development process when we can add one extra line to the job to do the extraction just prior to pushing to Transifex.

transifex/push.py
amangano-edx: This script is run using python3 on jenkins. Is it expected to be backwards compatible with python2 (I haven't tested that)? By using `python` here, are we communicating that the script should work with both python versions?

amangano-edx:
Replaced by https://github.com/edx/ecommerce-scripts/pull/61

PR #59, opened 2017-06-19
Added course_run and program tags for data sent to Sailthru

clintonb:
- All course runs now include the course_run tag 
- Programs now include a program tag and a tag with the slugified program type (e.g. program-micromasters)

### DIFF
--- a/edx_sailthru/sailthru_content/services/sailthru_translation_service.py
+++ b/edx_sailthru/sailthru_content/services/sailthru_translation_service.py
@@ -5,6 +5,8 @@ from __future__ import unicode_literals
 import datetime
 import logging
 
+from slugify import slugify
+
 from .constants import GFA_COURSE_RUN_LIST
 
 REQUIRED_SEATS_TYPES = ['verified', 'professional']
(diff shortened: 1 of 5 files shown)


clintonb:
FYI @efagin

edx_sailthru/sailthru_content/services/sailthru_translation_service.py
```diff
@@ -137,12 +139,13 @@ def translate_course_run(self, course_run, course, program_dictionary=None):
             sailthru_content['images'] = {'thumb': {'url': course_run['image']['src']}}
 
         # create the interest tags
-        tags = self._get_tags_from_property(course.get('subjects'), 'name', 'subject')
+        tags = ['course-run']
+        tags.extend(self._get_tags_from_property(course.get('subjects'), 'name', 'subject'))
         tags.extend(self._get_tags_from_property(course.get('owners'), 'key', 'school'))
         tags.extend(self._get_tags_from_property(course.get('sponsors'), 'key', 'school'))
 
         if len(tags) > 0:
```
efagin: Nit: there is no longer a way for this condition to be `false` with the new code.

PR #60, opened 2017-06-21
Add support for custom merge methods

amangano-edx:
These changes enable us to use a custom merge method when merging PRs produced via the translation scripts. This is necessary since several of our repositories (credentials, course-discovery, ecommerce, edx-mktg) do not allow merge commits.

Depends on https://github.com/edx/ecommerce-scripts/pull/55

JIRA: https://openedx.atlassian.net/browse/LEARNER-1089 

Replaces https://github.com/edx/ecommerce-scripts/pull/56

PR #61, opened 2017-06-21
Add Transifex push script

amangano-edx:
Adds script to extract translations strings and push them to Transifex.

Depends on https://github.com/edx/ecommerce-scripts/pull/60 and https://github.com/edx/ecommerce-scripts/pull/55

Replaces https://github.com/edx/ecommerce-scripts/pull/57

### DIFF
--- a/transifex/pull.py
+++ b/transifex/pull.py
@@ -23,13 +23,20 @@ from argparse import ArgumentParser
 from utils import DEFAULT_MERGE_METHOD, MERGE_METHODS, logger, repo_context
 
 
+# The name of the branch to use.
+BRANCH_NAME = 'update-translations'
+
+# The commit message to use.
+MESSAGE = 'Update translations'
+
+
 def pull(clone_url, merge_method=None, skip_compilemessages=False):
     """Pulls translations for the given repo.
 
     If applicable, commits them, pushes them to GitHub, opens a PR, waits for
     status checks to pass, then merges the PR and deletes the branch.
     """
-    with repo_context(clone_url, merge_method=merge_method) as repo:
+    with repo_context(clone_url, BRANCH_NAME, MESSAGE, merge_method=merge_method) as repo:
         logger.info('Pulling translations for [%s].', repo.name)
 
         repo.pull_translations()
--- a/transifex/push.py
+++ b/transifex/push.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+"""
+This script can be used to automatically extract translation strings,
+commit, push, and merge them to their respective repos, and then push them to Transifex.
+
+To use, export an environment variable `GITHUB_ACCESS_TOKEN`. The token requires
+GitHub's "repo" scope.
+
+Run the script from the root of this repo.
+
+    python transifex/push.py git@github.com:edx/course-discovery.git
+
+If you want to use a custom merge method pass the --merge-method option.
+
+    python transifex/push.py git@github.com:edx/course-discovery.git --merge-method rebase
+"""
+from argparse import ArgumentParser
+
+from utils import DEFAULT_MERGE_METHOD, MERGE_METHODS, logger, repo_context
+
+# The name of the branch to use.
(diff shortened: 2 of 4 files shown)


transifex/pull.py
```diff
@@ -6,238 +6,55 @@
 To use, export an environment variable `GITHUB_ACCESS_TOKEN`. The token requires
 GitHub's "repo" scope.
 
-Run the script from the root of this repo. To use the repo clone URLs stored in
-settings.yaml, simply run:
+Run the script from the root of this repo.
 
-    python transifex/pull.py
+    python transifex/pull.py git@github.com:edx/course-discovery.git
```
BenjiLee: Why is this hardcoded to the course-discovery repo?

BenjiLee:
Will review #55 and #60 before this PR.

transifex/push.py
```diff
@@ -0,0 +1,59 @@
...
+"""
+This script can be used to automatically extract translation strings,
+commit, push, and merge them to their respective repos, and then push them to transifex.
+
+To use, export an environment variable `GITHUB_ACCESS_TOKEN`. The token requires
+GitHub's "repo" scope.
+
+Run the script from the root of this repo.
+
+    python3 transifex/push.py git@github.com:edx/course-discovery.git
```
rlucioni: Following up on https://github.com/edx/ecommerce-scripts/pull/57#discussion_r123099995: you've already denoted the expected version of Python with the shebang line above. You'll typically see the expected version of Python aliased to `python` in most virtualenvs. Do whatever you're comfortable with here, but leaving it as `python` is sufficient.

transifex/push.py
```diff
@@ -0,0 +1,59 @@
...
+
+If you want to use a custom merge method pass the --merge-method option.
+
+    python3 transifex/push.py git@github.com:edx/course-discovery.git --merge-method rebase
+"""
+from argparse import ArgumentParser
+from utils import DEFAULT_MERGE_METHOD, MERGE_METHODS, logger, repo_context
```
rlucioni: Minor, but please group your imports correctly. isort can help if you're not sure.

transifex/utils.py
```diff
@@ -0,0 +1,342 @@
...
+                        self.name, self.pr.number, retries, self.MAX_MERGE_RETRIES
+                    )
+
+                    retries += 1
+
+                    if retries <= self.MAX_MERGE_RETRIES:
+                        # Poll every 5 minutes
+                        time.sleep(60 * 5)
+        else:
+            logger.info(
```
bderusha: Should this be louder than just `info`?  If the retries have failed it's possible that github is down and the comment might also fail.

transifex/push.py
```diff
@@ -0,0 +1,59 @@
...
+
+# The name of the branch to use.
+BRANCH_NAME = 'update-translation-strings'
+
+# The commit message to use.
+MESSAGE = 'Update translation strings'
+
+
+def push(clone_url, merge_method=None):
+    """Extracts translations for the given repo, commits them, pushes them to GitHub, opens a PR, waits for status
```
BenjiLee: When does this method merge and delete the PR?

transifex/utils.py
amangano-edx: I would prefer not to address this here, as this is one of several issues I noticed with the existing script that I decided not to do anything about during this pass of refactoring. I have created issues on the repository for each of the problems I identified that I would like to see fixed going forward. See https://github.com/edx/ecommerce-scripts/issues

transifex/push.py
amangano-edx: Merge step: https://github.com/edx/ecommerce-scripts/blob/e9f5a87cca02b7f8d782f755945730b9a7413bb5/transifex/push.py#L37
Delete PR step: https://github.com/edx/ecommerce-scripts/blob/e9f5a87cca02b7f8d782f755945730b9a7413bb5/transifex/utils.py#L69

PR #67, opened 2017-08-15
Fix course_type in sailthru

tasawernawaz:
[LEARNER-2244](https://openedx.atlassian.net/browse/LEARNER-2244)

### DIFF
--- a/edx_sailthru/sailthru_content/services/catalog_api_service.py
+++ b/edx_sailthru/sailthru_content/services/catalog_api_service.py
@@ -78,7 +78,8 @@ class CatalogApiService(object):
                 self.api_client.programs(),
                 page=1,
                 page_size=PROGRAMS_PAGE_SIZE,
-                marketable_course_runs_only=1
+                marketable_course_runs_only=1,
+                marketable=1
             )
             for program in program_array:
                 self._programs_dictionary[program['uuid']] = program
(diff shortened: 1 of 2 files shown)


edx_sailthru/sailthru_content/services/catalog_api_service.py
```diff
@@ -81,5 +81,6 @@ def get_program_dictionary(self):
                 marketable_course_runs_only=1
             )
             for program in program_array:
-                self._programs_dictionary[program['uuid']] = program
+                if program['status'] == 'active':
```
schenedx: Instead of manually filter, maybe you should just call with a flag documented at https://github.com/edx/course-discovery/blob/master/course_discovery/apps/api/v1/views/programs.py#L49

tasawernawaz:
@schenedx Code updated according to your suggestion.

edx_sailthru/sailthru_content/services/sailthru_translation_service.py
```diff
@@ -87,10 +87,10 @@ def _create_course_vars(self, course, course_run, url, site_name, program_dictio
             sailthru_content_vars['course_type'] = 'gfa'
 
         # If the course runs have associated programs, also append programs information
-        course_programs = course.get('programs')
-        if course_programs and program_dictionary:
+        course_run_programs = course_run.get('programs')
```
tasawernawaz: I don't think. Because API should be tested in course-discovery, which I believe already tested.

PR #68, opened 2017-09-06
add activation codes to existing users

tasawernawaz:
[LEARNER-2287](https://openedx.atlassian.net/browse/LEARNER-2287)

Please review @edx/learner-spartans

### DIFF
--- a/edx_sailthru/import_users_activation_codes/import_activation_codes.py
+++ b/edx_sailthru/import_users_activation_codes/import_activation_codes.py
@@ -1,0 +1,136 @@
+import argparse
+import logging
+
+import MySQLdb
+import os
+import pandas
+import sys
+from sailthru.sailthru_client import SailthruClient
+
+logger = logging.getLogger(__name__)
+
+
+def setup_logging():
+    logger.setLevel(logging.INFO)
+    logging.basicConfig(format='%(asctime)s - %(levelname)s - %(message)s')
+
+
+def get_list_of_inactive_users(args):
+    edxapp_db = MySQLdb.connect(host=args.mysql_host,
+                                user=args.mysql_user,
+                                passwd=args.mysql_pass,
+                                db=args.db)
+
+    if args.is_active is None:
+        active_query = ""
+    elif args.is_active:
+        active_query = "AND au.is_active=1"
+    else:
+        active_query = "AND au.is_active=0"
+
+    sql = """SELECT au.email, au.is_active, ar.activation_key
+    FROM auth_user au
+    INNER JOIN auth_registration ar
+    ON au.id=ar.user_id
+    """ + active_query
+
+    users_data = pandas.read_sql_query(sql, edxapp_db)
+
+    return users_data
+
+
+def update_user_on_sailthru(users_data, client):
+    logger.info("Starting updating users on sailthru")
+
+    for index, user in users_data.iterrows():
+        request_data = {
+            "id": user["email"],
+            "vars": {
+                "activation_key": user["activation_key"]
+            },
+
+        }
+        response = client.api_post('user', request_data)
+
+        if response.is_ok():
+            logger.info("User updated: %s" % user["email"])
+
+        else:
+            error = response.get_error()
+            logger.error("Error: " + error.get_message())
+            logger.error("Status Code: " + str(response.get_status_code()))
+            logger.error("Error Code: " + str(error.get_error_code()))
+
+
+def get_args(argv):
+    parser = argparse.ArgumentParser(description='Import activation codes of inactive users from LMS to Sailthru')
+
+    parser.add_argument(
+        '--sailthru_key',
+        help='Sailthru access key.'
+    )
+
+    parser.add_argument(
+        '--sailthru_secret',
+        help='Sailthru access secret.'
+    )
+
+    parser.add_argument(
+            '--mysql_host',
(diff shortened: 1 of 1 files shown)


edx_sailthru/import_users_activation_codes/import_activation_codes.py
```diff
@@ -0,0 +1,114 @@
...
+
+    parser.add_argument(
+        '--sailthru_secret',
+        help='Sailthru access secret.'
+    )
+
+    parser.add_argument(
+            '--mysql_host',
+            default=os.environ.get('EDXAPP_DB_HOST', ''),
+            help='Host address for ecommerce db replica.'
```
schenedx: Why do we want ecommerce db? The help text needs update?

edx_sailthru/import_users_activation_codes/import_activation_codes.py
```diff
@@ -0,0 +1,114 @@
...
+        )
+
+    parser.add_argument(
+        '--mysql_user',
+        default=os.environ.get('EDXAPP_DB_USER', ''),
+        help='File where result is saved.'
```
schenedx: Help text needs update

edx_sailthru/import_users_activation_codes/import_activation_codes.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+
+    parser.add_argument(
+        '--mysql_pass',
+        default=os.environ.get('EDXAPP_DB_PASSWORD', ''),
+        help='Password for ecommerce db replica.'
```
schenedx: Again, help text needs update

edx_sailthru/import_users_activation_codes/import_activation_codes.py
```diff
@@ -0,0 +1,114 @@
...
+        )
+
+    parser.add_argument(
+            '--db',
+            default=os.environ.get('EDXAPP_DB_NAME', ''),
+            help='File where course content is saved.'
```
schenedx: help text needs update

edx_sailthru/import_users_activation_codes/import_activation_codes.py
```diff
@@ -0,0 +1,114 @@
...
+
+        else:
+            error = response.get_error()
+            logger.error("Error: " + error.get_message())
+            logger.error("Status Code: " + str(response.get_status_code()))
+            logger.error("Error Code: " + str(error.get_error_code()))
+
+
+def get_args(argv):
+    parser = argparse.ArgumentParser(description='Sailthru import activation codes of users script')
```
schenedx: `Import activation codes of inactive users from LMS to Sailthru`

edx_sailthru/import_users_activation_codes/import_activation_codes.py
```diff
@@ -0,0 +1,114 @@
...
+            help='File where course content is saved.'
+    )
+
+    return parser.parse_args()
+
+
+def main(argv):
+    args = get_args(argv)
+
+    user_data = get_list_of_inactive_users(args)
```
schenedx: Can we be a bit more programmable on whether we should get active user or inactive users?
Say we can pass in an optional parameter `is_active` into this function.
If `is_active` is `True`, get all the active users
If `is_active` is `False`, get all the inactive users
If `is_active` is `None`, get all the users regardless
Then expose a script option that controls the `is_active` flag?
This suggestion is optional. I'll leave up to you to see if this is needed.

tasawernawaz: Updated according to your suggestion. Please review.

edx_sailthru/import_users_activation_codes/import_activation_codes.py
```diff
@@ -0,0 +1,136 @@
...
+
+    logger.info("Total inactive users are: %s." %len(user_data))
```
schenedx: Update this log message please. Maybe just log out `Total users are <number>`

edx_sailthru/import_users_activation_codes/import_activation_codes.py
```diff
@@ -0,0 +1,136 @@
...
+
+    sailthru_client = SailthruClient(args.sailthru_key, args.sailthru_secret)
+    update_user_on_sailthru(users_data=user_data, client=sailthru_client)
+
+    logger.info("All inactive users updated!")
```
schenedx: Update this log message to remove `inactive`

edx_sailthru/import_users_activation_codes/import_activation_codes.py
```diff
@@ -0,0 +1,136 @@
...
+
+    users_data = pandas.read_sql_query(sql, edxapp_db)
+
+    return users_data
+
+
+def update_user_on_sailthru(users_data, client):
+    logger.info("Starting updating users on sailthru")
+
+    for index, user in users_data.iterrows():
```
schenedx: Have you tried to use their batch/job interface recommended in the ticket?
you will be making millions of API calls this way. That's quite inefficient.

tasawernawaz: I read the docs related to it and it is stated that `If you want to apply the same update to all of the users specified, you should pass the update parameter.` So I believe we are updating every user with his/her own `activation_code` so we can do a bulk update.

PR #70, opened 2017-10-20
Modify 'site_name' string to include all course owners

uzairr:
### Description
Modify 'site_name' string to include all course owners

'site_name' in the 'course_weekly' data feeds on sailthru contains
only one course owner.In order to give credit to all organisations
of a given course there is a need to update 'site_name' in a way
that it will contain all the course owners.

### Testing
- [x] Unit

### DIFF
--- a/edx_sailthru/sailthru_content/services/sailthru_translation_service.py
+++ b/edx_sailthru/sailthru_content/services/sailthru_translation_service.py
@@ -110,6 +110,13 @@ class SailthruTranslationService(object):
         """ Return dict of sku for a course run. """
         return {seat['type']: seat['sku'] for seat in seats if seat['type'] in REQUIRED_SEATS_TYPES}
 
+    def get_course_owners(self, course):
+        owners = []
+        for owner in course.get('owners'):
+            #replace underscore in the owner's name if any
+            owners.append(owner['key'].replace('_', ' '))
+        return ' '.join(owners)
+
     def translate_course_run(self, course_run, course, program_dictionary=None):
         # get marketing url
         url = course_run.get('marketing_url', course.get('marketing_url'))
@@ -124,9 +131,9 @@ class SailthruTranslationService(object):
         if course_run.get('short_description'):
             sailthru_content['description'] = course_run.get('short_description')
 
-        # get first owner
+        # get owners
         if course.get('owners') and len(course.get('owners')) > 0:
-            sailthru_content['site_name'] = course['owners'][0]['key'].replace('_', ' ')
+            sailthru_content['site_name'] = self.get_course_owners(course)
 
         # use enrollment_end for sailthru expire_date
         if course_run.get('enrollment_end'):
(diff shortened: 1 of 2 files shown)


edx_sailthru/sailthru_content/services/sailthru_translation_service.py
```diff
@@ -110,6 +110,15 @@ def _return_sku(self, seats):
         """ Return dict of sku for a course run. """
         return {seat['type']: seat['sku'] for seat in seats if seat['type'] in REQUIRED_SEATS_TYPES}
 
+    def build_course_owners(self, course):
```
tasawernawaz: I believe we need unit test of this.

edx_sailthru/sailthru_content/services/sailthru_translation_service.py
```diff
@@ -126,7 +135,7 @@ def translate_course_run(self, course_run, course, program_dictionary=None):
 
         # get first owner
         if course.get('owners') and len(course.get('owners')) > 0:
-            sailthru_content['site_name'] = course['owners'][0]['key'].replace('_', ' ')
+            sailthru_content['site_name'] = self.build_course_owners(course)
```
tasawernawaz: may be `get_course_owners` is a better name of this function.

edx_sailthru/sailthru_content/services/sailthru_translation_service.py
```diff
@@ -110,6 +110,15 @@ def _return_sku(self, seats):
         """ Return dict of sku for a course run. """
         return {seat['type']: seat['sku'] for seat in seats if seat['type'] in REQUIRED_SEATS_TYPES}
 
+    def build_course_owners(self, course):
+        owners_length = len(course.get('owners'))
+        counter = 0
+        owners = ''
+        while counter < owners_length:
```
tasawernawaz: I believe you can do something like this, which will be more cleaner.
```
owners = []
for owner in course.get('owners'):
    owners.append(owner['key'].replace('_', ' '))
 return ' '.join(owners)
```

adeelkhan: I agree with @tasawernawaz . I guess one should avoid using while/counter construct and use iterators

edx_sailthru/sailthru_content/services/sailthru_translation_service.py
```diff
@@ -110,6 +110,12 @@ def _return_sku(self, seats):
         """ Return dict of sku for a course run. """
         return {seat['type']: seat['sku'] for seat in seats if seat['type'] in REQUIRED_SEATS_TYPES}
 
+    def get_course_owners(self, course):
+        owners = []
+        for owner in course.get('owners'):
+            owners.append(owner['key'].replace('_', ' '))
```
schenedx: Can you comment why we want to replace `_` with spaces?

PR #71, opened 2017-10-30
LEARNER-2653 Modified transifex.push for edx-platform push job

Ayub-Khan:

**Description :** 
This pr is part of the process to automate edx-platform translations process. As the edx-platform does not create any code changes and translations are later merged when pulled. This code would have failed for edx-platform. i have updated it such that it follows pull request process if there are code changes if there are not any code changes it will simply push the translations to transifex. Now this will work for edx-platform job too. Where push does not request any pull request. 
Al though now it will send the same files to transifex in case of mktg and edx-platform once every day even if there was no change.

### DIFF
--- a/transifex/push.py
+++ b/transifex/push.py
@@ -32,9 +32,14 @@ def push(clone_url, merge_method=DEFAULT_MERGE_METHOD):
     with repo_context(clone_url, BRANCH_NAME, MESSAGE, merge_method=merge_method) as repo:
         logger.info('Extracting translations for [%s].', repo.name)
         repo.extract_translations()
-        repo.commit_push_and_open_pr()
 
-        if repo.pr and repo.merge_pr() and repo.pr.is_merged():
+        if repo.is_changed():
+            repo.commit_push_and_open_pr()
+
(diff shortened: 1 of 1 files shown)


transifex/push.py
```diff
@@ -32,9 +32,14 @@ def push(clone_url, merge_method=DEFAULT_MERGE_METHOD):
...
         repo.extract_translations()
-        repo.commit_push_and_open_pr()
 
-        if repo.pr and repo.merge_pr() and repo.pr.is_merged():
+        if repo.is_changed():
+            repo.commit_push_and_open_pr()
+
+            if repo.pr and repo.merge_pr() and repo.pr.is_merged():
+                logger.info('Pushing translations to Transifex for [%s].', repo.name)
+                repo.push_translations()
```
amangano-edx: I'm not sure I understand the PR description, but if your goal was to have this push to transifex even when there are no code changes, I don't believe these changes would result in that happening. 

Could you please clarify what the intended behavior is here?

Ayub-Khan:
Closing this. As wont need this after https://github.com/edx/edx-platform/pull/16424.

PR #73, opened 2017-11-07
LEARNER-2653 Added owner as command line argument

Ayub-Khan:

**Description :** 
As edx-platform has no repo owner so changing logic to send a repo owner using the command line argument. This repo owner(git team) is more in terms of only translation jobs. Responsible for the success and failure of jobs.


transifex/requirements.txt
```diff
@@ -2,4 +2,4 @@ Django==1.11.2
 edx-i18n-tools==0.3.9
 PyGithub==1.29
 PyYAML==3.12
```
amangano-edx: Do we still need pyYAML?

Ayub-Khan: i will check if we are not using it in any other place i will remove it. we are no longer using it for openedx.yaml.

transifex/pull.py
```diff
@@ -73,6 +73,10 @@ def parse_arguments():
         help='URL to use to clone the repository.'
     )
     parser.add_argument(
+        'repo_owner',
+        help='git team responsible for translation jobs.'
```
amangano-edx: Nit: Maybe include that this is the user/team that will be pinged when errors occur?

transifex/push.py
```diff
@@ -46,6 +46,10 @@ def parse_arguments():
         help='URL to use to clone the repository.'
     )
     parser.add_argument(
+        'repo_owner',
+        help='git team responsible for translation jobs.'
```
amangano-edx: Nit: Maybe include that this is the user/team that will be pinged when errors occur?

amangano-edx:
Please be careful when you merge this as it relies on https://github.com/edx-ops/ecom-secure/pull/24#issuecomment-342587372 . I believe if you merge this before the other PR is ready, the translation jobs will break on Jenkins as they will not pass the new repo_owner argument.

transifex/requirements.txt
```diff
@@ -1,5 +1,8 @@
+Babel==1.3
 Django==1.11.2
+django-babel-underscore==0.5.2
 edx-i18n-tools==0.3.9
+mako==1.0.7
```
amangano-edx: Please update these requirements in a separate PR. This is not related to the original purpose of this PR, which was to simply add the --repo-owner flag. It should be reviewed and tested separately.

PR #74, opened 2017-11-15
LEARNER-2653 updates for edx-platform jobs

Ayub-Khan:

**Description :** 
-created and moved existed requiremts to requirements folder.
-add edx-platform specific requirements to requirements folder.
-renamed the existing requirements to default.
-updated edx-i18n-tools to latest.
-updated wait time because edx-platform takes more time any other repo.

### DIFF
--- a/transifex/requirements.txt
+++ b/transifex/requirements.txt
@@ -1,5 +1,5 @@
 Django==1.11.2
-edx-i18n-tools==0.3.9
+edx-i18n-tools==0.4.2
 PyGithub==1.29
 PyYAML==3.12
 transifex-client==0.12.4
--- a/transifex/requirements/edx-platform.txt
+++ b/transifex/requirements/edx-platform.txt
@@ -1,0 +1,11 @@
+Babel==1.3
+Django==1.8.18
+django-babel-underscore==0.5.2
+edx-i18n-tools==0.4.2
+mako==1.0.7
+PyGithub==1.29
+PyYAML==3.12
+transifex-client==0.12.4
+
+# Python libraries to be installed directly from github
+git+https://github.com/edx/django-wiki.git@b232cff63c5d342a5fe8f11a25f03803c7c45d65#egg=django-wiki
--- a/transifex/utils.py
+++ b/transifex/utils.py
@@ -114,10 +114,10 @@ edx = github.Github(github_access_token).get_organization('edx')
 
 
 class Repo:
-    # Make 9 attempts to merge the PR (the original attempt plus 8 retries), sleeping for 5 minutes between each
-    # attempt. This should result in a total 40 minutes of sleep time. Status checks should almost always
+    # Make 18 attempts to merge the PR (the original attempt plus 17 retries), sleeping for 5 minutes between each
+    # attempt. This should result in a total 90 minutes of sleep time. Status checks should almost always
     # complete in this period.
-    MAX_MERGE_RETRIES = 8
+    MAX_MERGE_RETRIES = 17
 
     """Utility representing a Git repo."""
     def __init__(self, clone_url, repo_owner, branch_name, message, merge_method=DEFAULT_MERGE_METHOD):


Ayub-Khan:
@amangano-edx kindly review this is the final version of this .

Ayub-Khan:
I will be updating this one more time to change edx django wiki branch to edx_release bcz currently its installing my branch with a minor code refactor.

transifex/utils.py
```diff
@@ -117,7 +117,7 @@ class Repo:
     # Make 9 attempts to merge the PR (the original attempt plus 8 retries), sleeping for 5 minutes between each
```
bderusha: Update comment.  It now makes 13 attempts

bderusha: Any reason why we needed to bump this?

transifex/requirements.txt
```diff
@@ -1,5 +1,11 @@
-Django==1.11.2
-edx-i18n-tools==0.3.9
+Babel==1.3
+Django==1.10.7
+django-babel-underscore==0.5.2
+edx-i18n-tools==0.4.1
```
bderusha: I'm assuming that this will also change re: https://github.com/edx/i18n-tools/pull/63/files

transifex/utils.py
amangano-edx: @bderusha my guess is that edx-platform takes a lot longer to run jenkins/CI checks than most of our other repos.

transifex/requirements.txt
```diff
@@ -1,5 +1,11 @@
-Django==1.11.2
-edx-i18n-tools==0.3.9
+Babel==1.3
+Django==1.10.7
```
bderusha: I talked with @amangano-edx about this a bit.  If there are edx-platform specific requirements it may make sense to refactor them into their own requirements file so we don't downgrade all of the systems versions.   Most IDAs will be ahead of edx-platform.  If they start upgrading to Django 2.0 who knows what could stop working when we decide to run them using 1.10.

For now I think it would be better to pull out edx-platform specific requirements into their own file.

transifex/utils.py
Ayub-Khan: yes i tried it with existing and it was not able to merge because checks took too mucch time.

bderusha:
@Ayub-Khan You will also need to update https://github.com/edx/jenkins-job-dsl-internal/blob/master/resources/translation-job.sh to use the new requirements files

Ayub-Khan:
ok

transifex/utils.py
```diff
@@ -114,10 +114,10 @@ def merge(self, commit_message=github.GithubObject.NotSet, merge_method=None):
 
 
 class Repo:
-    # Make 9 attempts to merge the PR (the original attempt plus 8 retries), sleeping for 5 minutes between each
-    # attempt. This should result in a total 40 minutes of sleep time. Status checks should almost always
+    # Make 13 attempts to merge the PR (the original attempt plus 12 retries), sleeping for 5 minutes between each
+    # attempt. This should result in a total 60 minutes of sleep time. Status checks should almost always
```
amangano-edx: Just a suggestion, but 60 minutes may not be enough time for edx-platform as the tests can take a long time. It might be good, rather than bumping this number for all consumers of the script, to instead allow for a command line argument so edx-platform can set the number much higher if necessary.

transifex/requirements/edx-platform.txt
```diff
@@ -0,0 +1,11 @@
+Babel==1.3
+Django==1.10.7
```
amangano-edx: This should probably be the same version of Django used by edx-platform, which appears to be 1.8.18.

transifex/requirements/edx-platform.txt
```diff
@@ -0,0 +1,11 @@
...
+django-babel-underscore==0.5.2
+edx-i18n-tools==0.4.2
+mako==1.0.7
+PyGithub==1.29
+PyYAML==3.12
+transifex-client==0.12.4
+
+# Python libraries to be installed directly from github
+git+https://github.com/edx/django-wiki.git@b232cff63c5d342a5fe8f11a25f03803c7c45d65#egg=django-wiki
```
amangano-edx: Can you clarify what this is needed for?

Ayub-Khan: Yes this includes a code refactor.

transifex/requirements/edx-platform.txt
Ayub-Khan: ok

transifex/utils.py
Ayub-Khan: I will incraese the number so that overall time is 90 mins. I think its ok to bump the number for all consumers because if they finish early process will be terminated so process will always be taking same time for them.

transifex/requirements/edx-platform.txt
Ayub-Khan: Yes this was to fix a import error.

PR #78, opened 2017-12-08
LEARNER-2653 Travis checks always required to pass before merge

Ayub-Khan:
Travis checks always required to pass before merge.
As the travis checks are optional for edx-platform repo so transifex bot do not wait for checks to pass. 
example https://github.com/edx/edx-platform/pull/16803
This pr will enforce the bot to always wait for the checks to pass before merging the pull request. 

Removed travis condition not all tests status.context contain travis in it. 

Changes have been tested localy.
FYI @feanil

### DIFF
--- a/transifex/utils.py
+++ b/transifex/utils.py
@@ -271,53 +271,50 @@ class Repo:
 
         retries = 0
         while retries <= self.MAX_MERGE_RETRIES:
-            try:
-                self.pr.merge(merge_method=self.merge_method)
-                logger.info('Merged [%s/#%d].', self.name, self.pr.number)
-                return True
-            except github.GithubException as e:
-                # Assumes only one commit is present on the PR.
-                statuses = self.pr.get_commits()[0].get_statuses()
+            # Assumes only one commit is present on the PR.
+            status = self.pr.get_commits()[0].get_combined_status()
 
-                # Check for any failing Travis builds. If any are found, notify the team
-                # and move on.
-                if any('travis' in s.context and s.state == 'failure' for s in statuses):
-                    logger.info(
-                        'A failing Travis build prevents [%s/#%d] from being merged. Notifying %s.',
-                        self.name, self.pr.number, self.owner
-                    )
-
-                    self.pr.create_issue_comment(
-                        '@{owner} a failed Travis build prevented this PR from being automatically merged.'.format(
-                            owner=self.owner
-                        )
-                    )
-
-                    return False
-                else:
-                    retries += 1
-
-                    if retries <= self.MAX_MERGE_RETRIES:
-                        logger.info(
-                            'Status checks on [%s/#%d] are pending. This is retry [%d] of [%d].',
-                            self.name, self.pr.number, retries, self.MAX_MERGE_RETRIES
-                        )
-
-                        # Poll every 5 minutes
-                        time.sleep(60 * 5)
-        else:
-            logger.info(
-                'Retry limit hit for [%s/#%d]. Notifying %s.',
-                self.name, self.pr.number, self.owner
-            )
-
-            # Retry limit hit. Notify the team and move on.
-            self.pr.create_issue_comment(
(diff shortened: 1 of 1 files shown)


Ayub-Khan:
This pull request is ready for review.

transifex/utils.py
```diff
@@ -271,53 +271,57 @@ def merge_pr(self):
...
-                return True
-            except github.GithubException as e:
-                # Assumes only one commit is present on the PR.
-                statuses = self.pr.get_commits()[0].get_statuses()
-
-                # Check for any failing Travis builds. If any are found, notify the team
-                # and move on.
-                if any('travis' in s.context and s.state == 'failure' for s in statuses):
+            # Assumes only one commit is present on the PR.
+            statuses = self.pr.get_commits()[0].get_statuses()
```
amangano-edx: Will this be different for edx-platform since it uses jenkins instead of travis?

transifex/utils.py
```diff
@@ -271,53 +271,57 @@ def merge_pr(self):
...
+            for s in statuses:
+                logger.info('[%s] status has status [%s]', s, s.state)
+            if any(s.state == 'failure' for s in statuses):
```
amangano-edx: Do we still need to check for `'travis in s.context'`, as in https://github.com/edx/ecommerce-scripts/pull/78/files#diff-5f7e2de26d943a8474540224f60a064aL284 ?

transifex/utils.py
```diff
@@ -271,53 +271,57 @@ def merge_pr(self):
...
-                # Assumes only one commit is present on the PR.
-                statuses = self.pr.get_commits()[0].get_statuses()
-
-                # Check for any failing Travis builds. If any are found, notify the team
-                # and move on.
-                if any('travis' in s.context and s.state == 'failure' for s in statuses):
+            # Assumes only one commit is present on the PR.
+            statuses = self.pr.get_commits()[0].get_statuses()
+            for s in statuses:
+                logger.info('[%s] status has status [%s]', s, s.state)
```
amangano-edx: Nit: you probably only need: `'[%s] has status [%s]'`

transifex/utils.py
```diff
@@ -271,53 +271,57 @@ def merge_pr(self):
...
+                if retries <= self.MAX_MERGE_RETRIES:
+                    logger.info(
+                        'Status checks on [%s/#%d] are pending. This is retry [%d] of [%d].',
+                        self.name, self.pr.number, retries, self.MAX_MERGE_RETRIES
+                    )
+
+                    # Poll every 5 minutes
+                    time.sleep(60 * 5)
+
+                else:
```
amangano-edx: Does this need to be nested within the `while` loop? I think it might make more sense for it to be outside the loop, as it was before, since the loop itself tests that `retries <= self.MAX_MERGE_RETRIES`

transifex/utils.py
Ayub-Khan: actually i removed that too because even all tests pass this api call returns some pending which stop bot from merging the pr. (https://api.github.com/repos/edx/edx-platform/statuses/24ec64615c8ce7751d89cf0f828554ff0843afb8)
Started using the get combined status method which returns one status and makes the following api call (https://api.github.com/repos/edx/edx-platform/statuses/24ec64615c8ce7751d89cf0f828554ff0843afb8)

transifex/utils.py
Ayub-Khan: during my debugging of edx-platform job none of the job context had travis in it. also that could be changed without our knowledge by devops, so removed that condition. now it checks for the status of all tests running and when the final state is success it merges.

transifex/utils.py
Ayub-Khan: sorry this was only for testing purposes removed this condition.

transifex/utils.py
Ayub-Khan: moved it outside the while loop.

PR #80, opened 2018-01-22
Argumrnt added to remove git process for edx-platform

Ayub-Khan:
As the make extract creates a lot of changes for edx-platform this will prevent push translations job to push those translations to edx-platform repo.

### DIFF
--- a/transifex/push.py
+++ b/transifex/push.py
@@ -25,7 +25,7 @@ BRANCH_NAME = 'update-translation-strings'
 MESSAGE = 'Update translation strings'
 
 
-def push(clone_url, repo_owner, merge_method=DEFAULT_MERGE_METHOD, no_commit=False):
+def push(clone_url, repo_owner, merge_method=DEFAULT_MERGE_METHOD, skip_commit=False):
     """Extracts translations for the given repo, commits them, pushes them to GitHub, opens a PR, waits for status
         checks to pass, merges the PR, deletes the branch, and pushes the updated translation files to Transifex.
     """
@@ -33,7 +33,7 @@ def push(clone_url, repo_owner, merge_method=DEFAULT_MERGE_METHOD, no_commit=Fal
         logger.info('Extracting translations for [%s].', repo.name)
         repo.extract_translations()
 
-        if not no_commit:
+        if not skip_commit:
             repo.commit_push_and_open_pr()
             if repo.pr and repo.merge_pr() and repo.pr.is_merged():
                 push_translations_to_transifex(repo)
@@ -45,7 +45,6 @@ def push_translations_to_transifex(repo):
     """
     Calls the push translations make target to push new translation strings to Transifex.
     """
-    logger.info('Pushing translations to Transifex for [%s].', repo.name)
     repo.push_translations()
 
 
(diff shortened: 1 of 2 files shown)


amangano-edx:
Does this actually solve the problem? Wouldn't we still push the .po files with the extra changes we're concerned about to transifex?

Ayub-Khan:
@amangano-edx  This does not solve the problem we are facing but as solving the problem can take a lot of time. We can move forward with this to make push-pull translation jobs working to remove the manual process and create a new ticket to figure out why extract is making so many changes in the python 3. While automated jobs are working.

Ayub-Khan:
Also this is related to push translations only and this was suggested by you earlier i made a note to do it later so that we can move forward at that time and come back to it once we have working jobs. This has nothing to do with the pull translations.

amangano-edx:
@Ayub-Khan I don't think we should push this out until we understand the additional changes extract is making. Doing so could result in incorrect strings being pushed to transifex. Until we understand that better, we should continue extracting/pushing to transifex manually.

Ayub-Khan:
@amangano-edx  sure i will put those on hold. for the time being.

transifex/push.py
```diff
@@ -25,18 +25,28 @@
...
     """
     with repo_context(clone_url, repo_owner, BRANCH_NAME, MESSAGE, merge_method=merge_method) as repo:
         logger.info('Extracting translations for [%s].', repo.name)
         repo.extract_translations()
-        repo.commit_push_and_open_pr()
 
-        if repo.pr and repo.merge_pr() and repo.pr.is_merged():
-            logger.info('Pushing translations to Transifex for [%s].', repo.name)
-            repo.push_translations()
+        if not no_commit:
```
amangano-edx: Nit: I'd prefer to call this something positive, like skip_commit, as that's easier to read in if statements.

amangano-edx: Actually, no_commit is fine, but it'd be easier to read if we switched the if statetments around.
```
if no_commit:
  push_translations_to_transifex(repo)
else:
  # The other case...
```

transifex/push.py
```diff
@@ -25,18 +25,28 @@
...
+                push_translations_to_transifex(repo)
+        else:
+            push_translations_to_transifex(repo)
+
+
+def push_translations_to_transifex(repo):
+    """
+    Calls the push translations make target to push new translation strings to Transifex.
+    """
+    logger.info('Pushing translations to Transifex for [%s].', repo.name)
```
amangano-edx: Could you just add this log call to the `push_translations` method defined on `Repo`?

transifex/push.py
```diff
@@ -55,9 +65,14 @@ def parse_arguments():
         default=DEFAULT_MERGE_METHOD,
         help='Method to use when merging the PR. See https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button for details.'
     )
+    parser.add_argument(
+        '--no-commit',
+        action='store_true',
+        help='Use this if you do not want to commit changes to repo.'
```
amangano-edx: Do we need to set `default=False`?

amangano-edx:
These changes look fine, but before merging, could you please remind me why we don't want to commit the .po file changes that push_translations produces anymore?

transifex/push.py
Ayub-Khan: have changed to skip_commit.

PR #81, opened 2018-02-06
Updated Branch Name

Ayub-Khan:
A lot of Translation pull requests fail because
of already existing pull request or branch with same name.
updated it to use date to create branch.

### DIFF
--- a/transifex/utils.py
+++ b/transifex/utils.py
@@ -1,3 +1,4 @@
+import datetime
 import logging
 import os
 import re
@@ -135,7 +136,7 @@ class Repo:
 
         self.github_repo = edx.get_repo(self.name)
         self.owner = repo_owner
-        self.branch_name = branch_name
+        self.branch_name = branch_name + str(datetime.date.today())
         self.message = message
         self.pr = None
         self.merge_method = merge_method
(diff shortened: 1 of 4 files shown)


Ayub-Khan:
@edx/learner-spartans  kindly review
FYI @amangano-edx

amangano-edx:
My only concern is how we ensure that orphaned branches get cleaned up.

transifex/utils.py
```diff
@@ -321,7 +322,7 @@ def cleanup(self):
 
         If applicable, also deletes the merged branch from GitHub.
         """
-        if self.pr and self.pr.is_merged():
+        if self.pr:
             logger.info('Deleting merged branch %s:%s.', self.name, self.branch_name)
```
amangano-edx: This log message should be updated, as the branch might not have been merged.

amangano-edx: Do you need to do anything to close the PR first if it's still open before you delete the branch?

amangano-edx:
This means that the script will always delete the branch it's working with, even if it wasn't able to merge the PR. Does that make it more difficult for us to identify and fix problems that occur that are not transient-issues (Maybe Travis fails because the translations do not validate, for instance)?

Also, if someone manually creates a branch with the same name the script uses, the script may accidentally delete their branch. I'm not sure if that's a big problem or not, but it might make sense to make the branch_name more specific to the script to prevent possible issues.

transifex/utils.py
Ayub-Khan: ok. and no deleting branch automatically closes the pull request.

Ayub-Khan:
As script only creates branch after running all the commands successfuly so if it fails there wont be a branch on prod. No its gonna be same as before just that before we find issues on open pull request now we will receive for closed pull request.
The chance of some one manually creating a branch with the same name as the script uses is gonna be rare bcz it includes date. but i have modified it a bit more.

amangano-edx:
Please remember to squash your commits.

PR #83, opened 2018-02-21
Attempt to log Github error on PR

efischer19:
As demonstrated [here](https://github.com/edx/studio-frontend/pull/131#issuecomment-367402198)

For comparison, [here](https://tools-edx-jenkins.edx.org/job/translations/job/studio-frontend-pull_translations/10/console) is a run from last weekend that silently failed at this spot.

### DIFF
--- a/transifex/utils.py
+++ b/transifex/utils.py
@@ -297,6 +297,17 @@ class Repo:
                         'Failed to merge [%s/#%d], because of the exception [%s]',
                         self.name, self.pr.number, e,
                     )
+                    try:
+                        # If it's possible, a PR message would be appreciated in these cases.
+                        # Github may just be super borked though, hence the empty catch.
+                        self.pr.create_issue_comment(
+                            '@{owner} a github exception prevented this PR from being automatically merged:\n`{error}`'.format(
+                                owner=self.owner,
+                                error=e
+                            )
+                        )
+                    except:
+                        pass
                     return False
             else:
                 retries += 1


efischer19:
@amangano-edx, you've worked on this code before - mind reviewing this small addition when you get a chance?

We really want to keep the review requirement on studio-frontend PRs; this change will allow us to receive an email notification to re-open and merge translation update PRs (which will fail due to the "review required" stipulation) when we get in after the weekend.

transifex/utils.py
```diff
@@ -297,6 +297,17 @@ def merge_pr(self):
...
+                        # If it's possible, a PR message would be appreciated in these cases.
+                        # Github may just be super borked though, hence the empty catch.
+                        self.pr.create_issue_comment(
+                            '@{owner} a github exception prevented this PR from being automatically merged:\n`{error}`'.format(
+                                owner=self.owner,
+                                error=e
+                            )
+                        )
+                    except:
+                        pass
```
amangano-edx: A GithubException could occur because Github is down. In that case, leaving a PR comment would also fail. Furthermore, I'm not sure that opening the PR makes sense at all here since we know that it will always fail. I also think that the spartan team recently added logic to automatically remove the PR/branch from Github when the script cleans up, so it may disappear before you have a chance to merge it.

amangano-edx:
I'm not sure about this approach. If you really want to run this script, and know that the merge-attempts will always fail, you may want to consider adding a command line option to skip the merge and cleanup steps.

openedx-webhooks:
Although this pull request is already closed,  I'm still watching it for updates.

There is nothing you have to do. No action is needed from your side.  Thanks.

PR #85, opened 2018-02-28
LEARNER-4195 Make sure jenkins automated jobs fail when error during the job

Ayub-Khan:
This pull request is to make sure jenkins sailthru and translation jobs fail when a error occur during the job process. The only was to do that is to raise exception and break the program.

### DIFF
--- a/edx_sailthru/sailthru_content/services/sailthru_api_service.py
+++ b/edx_sailthru/sailthru_content/services/sailthru_api_service.py
@@ -23,6 +23,7 @@ class SailthruApiService(object):
                 response.json['error'],
                 response.json['errormsg']
             )
+            raise RuntimeError('Failed to connect with Sailthru API')
         else:
             for body in response.json['content']:
                 logging.info(body)
@@ -57,6 +58,7 @@ class SailthruApiService(object):
             logging.error("Error: " + error.get_message())
             logging.error("Status Code: " + str(response.get_status_code()))
             logging.error("Error Code: " + str(error.get_error_code()))
+            raise RuntimeError('Failed to connect with Sailthru API')
 
     def upload(self, library_items, report_email=None):
         if not library_items:
@@ -80,7 +82,7 @@ class SailthruApiService(object):
                     response.json['error'],
                     response.json['errormsg']
                 )
-                return
+                raise RuntimeError('Failed to connect with Sailthru API')
 
             sailthru_content = response.json['content']
             if not sailthru_content:
--- a/transifex/utils.py
+++ b/transifex/utils.py
@@ -276,7 +276,7 @@ class Repo:
             status = self.pr.get_commits()[0].get_combined_status()
 
             if status.state == 'failure':
-                logger.info(
+                logger.error(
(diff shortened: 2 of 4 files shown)


edx_sailthru/sailthru_content/services/sailthru_api_service.py
```diff
@@ -80,7 +82,7 @@ def clear(self):
                     response.json['error'],
                     response.json['errormsg']
                 )
-                return
+                raise RuntimeError('Failed to connect with Sailthru API')
```
amangano-edx: I'd recommend asking @schenedx to take a look at these sailthru changes, as I'm not familiar with these scripts.

transifex/utils.py
```diff
@@ -297,7 +300,8 @@ def merge_pr(self):
                         'Failed to merge [%s/#%d], because of the exception [%s]',
                         self.name, self.pr.number, e,
                     )
-                    return False
+                    self.cleanup()
```
amangano-edx: The cleanup method is called in a finally block, so I don't think you need to call it explicitly here, or anywhere else.

amangano-edx:
Could you please clarify the purpose of these changes? Is it to get the builds to appear as 'failed' on jenkins so that it's easier to see when things go wrong?

edx_sailthru/sailthru_content/services/sailthru_api_service.py
Ayub-Khan: sure.

Ayub-Khan:
@amangano-edx yes the purpose is to get builds appear as 'failed'. Currently build shows success this way build will show failed if job failed and it would be easy to identify the problem. Sailthru job was failing for around 2 months and we had no idea because builds were showing success and issue was later identified when we checked courses were missing from feed. Although it was optional for translations because we get git notification too if a job failed but i moved forward by failing those too to make them similar and to represent job failure too.

transifex/utils.py
```diff
@@ -297,7 +297,7 @@ def merge_pr(self):
                         'Failed to merge [%s/#%d], because of the exception [%s]',
                         self.name, self.pr.number, e,
                     )
-                    return False
+                    raise RuntimeError('Failed to merge.')
```
amangano-edx: It might be good to specify that the failure was due to a GithubException, and maybe include the exception message?

amangano-edx:
I think the translation script changes are probably fine, though there may be other places in the where they can fail but not raise an error. I'd recommend looking at 'push.py' and 'pull.py'.

I cannot speak to the Sailthru changes, as I'm not familiar with that code.

transifex/utils.py
Ayub-Khan: github exception is already being shown by logger.info so no need to show it again i think.

amangano-edx: It might be good to log that at the error level instead of info, so it's clear in the log that it's an error.

Ayub-Khan:
@amangano-edx  lets discuss this tonight

Ayub-Khan:
Merging this.

PR #94, opened 2018-09-20
Adds killswitch var to group headers

jlajoie:


cloudflare_worker/worker.js
```diff
@@ -102,15 +110,20 @@ function rolloutGroupHeaders(request){
...
 
-function presortToControl(request, control_cookie_pattern, control_group){
+function presortToControl(request, control_cookie_pattern, control_group, killswitch){
   let responseObj = {
     assignment: '',
     isNew: false,
   }
 
-  // If a user has a control cookie, keep them in the control group
+  // If the killswitch is enabled just return
```
bderusha: This code doesn't seem necessary.  If the switch is on it doesn't matter what this returns because we will never use this object.

jlajoie: :+1:

cloudflare_worker/worker.js
```diff
@@ -102,15 +110,19 @@ function rolloutGroupHeaders(request){
   return headers
 }
 
-function presortToControl(request, control_cookie_pattern, control_group){
+function presortToControl(request, control_cookie_pattern, control_group, killswitch){
+  // If the killswitch is enabled just return
+  if ( killswitch ) {
```
bderusha: Do we need this block at all?

jlajoie: Nope! Was just trying to save a few calls but it really doesn't matter, can remove and reduce params going to the function

cloudflare_worker/worker.js
```diff
@@ -62,9 +66,13 @@ function rolloutGroupHeaders(request){
   let isNew = false  // is the group newly-assigned?
 
   // Determine which group this request is in.
-  let controlAssignment = presortToControl(request, control_cookie_pattern, control_group_name)
+  let controlAssignment = presortToControl(request, control_cookie_pattern, control_group_name, killswitch)
```
bderusha: remove this param, then :shipit:

jlajoie: Added it back in, currently was setting the cookie a whole bunch of times when we don't need to do so if the killswitch is on and they already have a forced cookie.

PR #101, opened 2018-11-09
LEARNER-6636 Increased translation job wait time

Ayub-Khan:
LEARNER-6636 Increased translation job wait time to fix edx-mktg


Ayub-Khan:
This is not gonna fix the issue.

PR #116, opened 2019-01-28
new worker code for cs bz rollout 

bderusha:


bderusha:
**NOTE TO US:**   Do not merge this until it is LIVE in cloudflare

PR #118, opened 2019-02-14
update translation job commit message to be semantic

albemarle:


transifex/push.py
```diff
@@ -22,7 +22,7 @@
 BRANCH_NAME = 'transifex-bot-update-translation-strings'
 
 # The commit message to use.
-MESSAGE = 'Update translation strings'
+MESSAGE = 'fix(translations): update translation strings'
```
robrap: Nit: Maybe use `fix(i18n)` which is shorter?

PR #121, opened 2019-04-29
Fix translations jobs by adding constraint.txt

schenedx:
All the translation job are failing in jenkins. 
The root cause is that, on transifex-client 1.12.4, which this is pinned, the urllib3 dependency is not pinned. See https://github.com/transifex/transifex-client/blob/128b212a5f07d1a1475ebd10fda0bf6dbbe2c055/requirements.txt
Because of it, a new urllib3 (1.25.1) was installing and both transifex and requests libraries can't support that.
The added a constraint file to pin the urllib3 be < 1.25


schenedx:
@albemarle Understand that would be more structured. But that also means a change to the command line, which means a change to the jenkins job. I don't see values doing so much work for this.

PR #137, opened 2020-02-20
Adding django-statici18n support to pull translation job.

adeelkhan:
This patch would enable pull translation job for edx-platform
to generate aggregate JS Catalog for all apps
via django-static18n package. For now developer has to do
this process to enable js translations to be available in production.

PROD-960

### DIFF
--- a/transifex/pull.py
+++ b/transifex/pull.py
@@ -18,6 +18,9 @@ If you want to skip the compile messages step, pass the --skip-compilemessages o
 
     python transifex/pull.py git@github.com:edx/course-discovery.git --skip-compilemessages
 """
+import os
+import shutil
+
 from argparse import ArgumentParser
 
 from utils import DEFAULT_MERGE_METHOD, MERGE_METHODS, logger, repo_context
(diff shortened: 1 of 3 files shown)


transifex/pull.py
```diff
@@ -37,10 +50,17 @@ def pull(clone_url, repo_owner, merge_method=DEFAULT_MERGE_METHOD, skip_compilem
     If applicable, commits them, pushes them to GitHub, opens a PR, waits for
     status checks to pass, then merges the PR and deletes the branch.
     """
+    with repo_context(CONFIGURATION_REPO_URL, repo_owner, BRANCH_NAME, MESSAGE, merge_method=merge_method) as config_repo:
+        logger.info('Pulling lms/studio settings from [%s].', config_repo.name)
+
+        shutil.copy('./docker/build/edxapp/lms.yml', '../')
+        shutil.copy('./docker/build/edxapp/studio.yml', '../')
```
waheedahmed: Can't we give this path directly to those environment variables instead of copying them?

adeelkhan: Nope. This code is trying to pull lms/studio settings out of configuration repo to its parent directory. As the With context deletes the repo we have to move the settings to a place where the edx-platform repo could have access to them. So this code is just moving settings out of config repo and the above environment variables are accessing them from inside edx-platform repo. 
Directory structure is
ecommerce-script 
  |-> lms.yml (copied from configuration repo)
  |-> studio.yml (copied from configuration rep)
  |-> configuration (this would eventually be cleaned up when code is out of with context)
  |-> edx-platform

waheedahmed: Got it, thanks!

PR #139, opened 2020-04-02
Maxmind downloaded url updated

adeelehsan:
Maxmind has updated the downloading of
geoip database. Added these changes.

PROD-1329

### DIFF
--- a/geoipupdate/geoip.py
+++ b/geoipupdate/geoip.py
@@ -1,37 +1,40 @@
 import hashlib
 import os
 import urllib
-import gzip
+import tarfile
 import shutil
 import logging
 
 
-MAXMIND_URL = 'http://geolite.maxmind.com/download/geoip/database/'
-MAXMIND_FILENAME = 'GeoLite2-Country.mmdb.gz'
+license_key = os.environ['LICENSE-KEY']
+MAXMIND_URL = 'https://download.maxmind.com/app/geoip_download?edition_id=GeoLite2-Country' \
+              '&license_key={key}&suffix=tar.gz'.format(key=license_key)
+MAXMIND_SHA256_URL = 'https://download.maxmind.com/app/geoip_download?edition_id=GeoLite2-Country' \
+                     '&license_key={key}&suffix=tar.gz.sha256'.format(key=license_key)
 LOCAL_FILENAME = 'GeoLite2-Country.mmdb'
 GEOIP_PATH = "/common/static/data/geoip/"
 
 
-def match_md5(fp, md5_url):
+def match_sha256(_file, sha256_url):
     """
 
     Purpose of the method to make sure downloaded file is accurate.
 
     Args:
-         fp (file): downloaded file
-         md5_url (str): url to read md5 hash of file
+         _file (file): downloaded file
+         sha256_url (str): url to read sha256 hash of file
 
     Returns:
         bool: Returns True if downloaded file hash and hash returned by url
         are same otherwise returns False
 
     """
-    md5 = urllib.request.urlopen(md5_url).read()
-    m = hashlib.md5()
-    for line in fp:
-        m.update(line)
-    fp.seek(0)
-    return m.hexdigest() == md5.decode()
+    sha256 = urllib.request.urlopen(sha256_url).read()
+    hash_object = hashlib.sha256()
+    file_content = open(_file, 'rb').read()
+    hash_object.update(file_content)
+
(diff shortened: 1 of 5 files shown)


geoipupdate/geoip.py
```diff
@@ -1,37 +1,41 @@
...
 
 
-MAXMIND_URL = 'http://geolite.maxmind.com/download/geoip/database/'
-MAXMIND_FILENAME = 'GeoLite2-Country.mmdb.gz'
+license_key = os.environ['LICENSE-KEY']
+MAXMIND_URL = 'https://download.maxmind.com/app/geoip_download?edition_id=GeoLite2-Country' \
+              '&license_key={key}&suffix=tar.gz'.format(key=license_key)
+MAXMIND_SHA256_URL = 'https://download.maxmind.com/app/geoip_download?edition_id=GeoLite2-Country' \
+                     '&license_key={key}&suffix=tar.gz.sha256'.format(key=license_key)
+MAXMIND_FILENAME = 'GeoLite2-Country_20200324.tar.gz'
```
waheedahmed, line 14: Can we remove the date stamp from the file name? Because the date will not be the same for each run.

geoipupdate/geoip.py
```diff
@@ -1,37 +1,41 @@
...
 
     """
-    md5 = urllib.request.urlopen(md5_url).read()
-    m = hashlib.md5()
-    for line in fp:
-        m.update(line)
-    fp.seek(0)
-    return m.hexdigest() == md5.decode()
+    sha256 = urllib.request.urlopen(sha256_url).read()
+    m = hashlib.sha256()
```
adeelkhan, line 34: As m here is a hash object, please rename it to be more explicit eg hash_object

geoipupdate/geoip.py
```diff
@@ -1,37 +1,41 @@
...
+    _file = open(_file, 'rb').read()
+    m.update(_file)
```
adeelkhan, line 36: I would suggest `file_content = open(_file, 'rb').read() ` as you are generating hash on file content

geoipupdate/geoip.py
```diff
@@ -78,23 +82,24 @@ def download_file(current_directory):
...
+                    _file = outfile.extractfile(member)
+                    write(_file, existing_file)
+        os.remove(downloaded_file)
+    else:
+        logging.info("Downloaded file hash did't matched. Aborting the operation")
+        try:
+            os.remove(downloaded_file)
+        finally:
+            raise ValueError(
+                'md5 of %s doesn\'t match the signature.' % downloaded_file
```
adeelkhan, line 104: Change the log to to output **sha256.** . Plus also change the docstring https://github.com/edx/ecommerce-scripts/blob/master/geoipupdate/geoip.py#L64

geoipupdate/test/test_geoip_update.py
```diff
@@ -1,45 +1,51 @@
...
     @patch('urllib.request.urlopen')
-    def test_md5_mismatch(self, mock_urlopen='http://foo'):
-        mock_md5 = MagicMock()
-        mock_md5.getcode.return_value = 200
-        mock_md5.read.return_value = '4059D862F09C315470DC8FF0FF6EBCAF'.encode('utf-8')
-        mock_urlopen.return_value = mock_md5
+    def test_sha256_mismatch(self, mock_urlopen='http://foo'):
+        mock_sha256 = MagicMock()
+        mock_sha256.getcode.return_value = 200
+        mock_sha256.read.return_value = '4059D862F09C315470DC8FF0FF6EBCAF'.encode('utf-8')
```
adeelkhan, line 19: I think you need to change the return value incorporate sha256 hash. I think md5 and sha256 output hash would be different. This doc says one is 32 digits and other is 64 digits. https://www.mobilefish.com/services/hash_generator/hash_generator.php

geoipupdate/test/test_geoip_update.py
```diff
@@ -66,15 +74,13 @@ def test_file_content(self):
     def test_raise_value_error(self, mock_urlretrieve, mock_urlopen):
```
adeelkhan, line 74: Please remove `mock_urlretrieve` . Seems to be dead code.

adeelehsan: Its not dead code we are mocking the file download . Have a look https://github.com/edx/ecommerce-scripts/pull/139/files#diff-58f2ef0f311b67612cde6a98629b4a34R82

PR #153, opened 2021-03-23
feat: drop sailthru scripts

mikix:
They were mostly one-time import scripts when setting sailthru up. Or GTM & drupal related.

The one script that was still actively used until recently was the sync_library script, which pushed discovery & program info into sailthru. But that is no longer needed either and was dropped in the below PR.

jenkins-job-dsl-internal side of this here: https://github.com/edx/jenkins-job-dsl-internal/pull/523


jhan217 (approved):

mikix:
OK, SREs rid of the jenkins job itself, and I have removed the job definition, so here's hoping. I'll land this now, nothing should be using these scripts.

PR #165, opened 2021-08-02
feat: separate edx-proctoring translation requirements

ashultz0:
edx-proctoring needs simple history now and that causes the translation job to fail https://tools-edx-jenkins.edx.org/job/translations/job/edx-proctoring-push_translations/35/ but we'd prefer not to add simple history  to every translation job

the better long-term fix is to update the proctoring translation targets to use i18n_tools


schenedx (approved):
👍 LGTM

PR #168, opened 2021-09-29
chore: add edx-analytics-dashboard python requirements to enable translation extraction

schenedx:
@edx/masters-devs-cosmonauts Please review

Without these requirements installed on the jenkins, the [translation push job](https://tools-edx-jenkins.edx.org/job/translations/job/edx-analytics-dashboard-push_translations/) will fail.


matthugs (approved):
looks like the next step is to update https://github.com/edx/edx-internal/blob/2ede526bcc544938548568184dfc8a9c45a48793/tools-edx-jenkins/translation-jobs.yml#L48

PR #184, opened 2022-03-25
build: add file for missing ora2 library permissions

matthugs:
following the pattern of how edx-proctoring is set up, motivated by an
error like `ModuleNotFoundError: No module named 'waffle'`

also tangentially updates the version of a library I'm concerned with to latest, which may impact translations done by this & other plugins

downstream: https://github.com/edx/edx-internal/pull/6359


schenedx (approved):
👍

mattcarter (approved):

PR #191, opened 2022-05-16
fix: Fix translations automerge for repos with long-running status checks

timmc-edx:
`get_combined_status` was always returning "success", so the script would
try to merge edx-platform PRs too quickly. This change uses the partially
documented (but apparently more effective) `mergeable_state` on the PR
itself to get better results. (In practice, seems to return "blocked" for
both pending and failing commits, so just lump those together along with
all other failure modes.)

Also:

- Log the URL of the translation PR, for convenience.
- Delete existing repo before cloning to directory. This should help with
  recovering from an interrupted run -- otherwise, future attempts to clone will fail.
- Tighten up repo-name matching regex for security.

ref: 2U ARCHBOM-2123

### DIFF
--- a/transifex/utils.py
+++ b/transifex/utils.py
@@ -135,7 +135,7 @@ class Repo:
             path=parsed.path
         )
 
-        match = re.match(r'.*edx/(?P<name>.*).git', self.clone_url)
+        match = re.fullmatch(r'.*edx/(?P<name>[a-zA-Z0-9_-]*)\.git', self.clone_url)
         self.name = match.group('name')
 
         self.github_repo = self._get_repo(self.name)
@@ -166,6 +166,8 @@ class Repo:
 
     def clone(self):
         """Clone the repo."""
+        # First clear out any preexisting stale checkout (e.g. from a previous, interrupted run)
+        subprocess.run(['rm', '-rf', '--', self.name], check=True)
         subprocess.run(['git', 'clone', '--depth', '1', self.clone_url], check=True)
 
     def branch(self):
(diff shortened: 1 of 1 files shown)


timmc-edx:
Successful run, using this branch: https://tools-edx-jenkins.edx.org/job/translations/job/edx-platform-pull_translations/382/console

transifex/utils.py
```diff
@@ -301,21 +304,14 @@ def merge_pr(self):
...
-            if state == 'failure':
-                logger.error(
-                    'A failing CI build prevents [%s/#%d] from being merged. Notifying %s.',
-                    self.name, self.pr.number, self.owner
-                )
+            # Update our local PR object from the server (undocumented in PyGithub?)
+            self.pr.update()
+            # Not fully documented in regular API, but this looks like the same thing in the GraphQL docs:
+            # https://docs.github.com/en/graphql/reference/enums#mergestatestatus
+            state = self.pr.mergeable_state
```
robrap, line 312: Consider opening issues around these questions or lacks, and pointing to the opened issues.

timmc-edx: I think it's OK to leave it as "it will work until it breaks" since this is just a way to avoid trying to merge too early; GitHub's own mergeability checks are the final backstop.

I've filed an issue on PyGithub for documenting the method, and I'll add links to that and to the Github discussion around their docs on `mergeable_state`, but I don't think we need to track it in our own issues.

timmc-edx: (In retrospect, I guess you weren't suggesting filing Jira tickets, heh.)

robrap: Correct. Just github issues where and if appropriate.

PR #192, opened 2022-05-27
feat: fix pull_translations for repos that require reviews

bseverino:
https://2u-internal.atlassian.net/browse/WS-2908

With the new check for mergeable_state, the job will always fail for
repos that require reviews, as it will return as 'blocked' whether or
not CI is passing. Use get_combined_status and get_check_runs to
determine status in this case.


bseverino:
Tested here: https://tools-edx-jenkins.edx.org/job/translations/job/prospectus-pull_translations/808

The job failed with this error, but the PR DID merge successfully (https://github.com/edx/prospectus/pull/4031), so I'm not sure what happened there:

```
Failed to merge [prospectus/#4031], because of the exception [405 {"message": "Pull Request is not mergeable", "documentation_url": "https://docs.github.com/rest/reference/pulls#merge-a-pull-request"}]
```

transifex/utils.py
```diff
@@ -340,6 +372,49 @@ def merge_pr(self):
...
+        commit = self.pr.get_commits().reversed[0]
+
+        total_statuses = commit.get_statuses().totalCount
+        if total_statuses > 0:
+            combined_status = commit.get_combined_status().state
+            logger.info(f'Combined status for {self.name}/#{self.pr.number}: {combined_status}')
+            if combined_status in [PENDING, FAILURE]:
+                return combined_status
+
+        # Check runs are not included in statuses, iterate over these before determining
```
srwang, line 398: what are check runs?

srwang (approved):
this looks great! 

just for my understanding, what is the new check for mergeable_state? where did it come from?

transifex/utils.py
bseverino: [commit statuses](https://docs.github.com/en/rest/commits/statuses) are generated by external services (e.g. codecov), while check runs use the [GH Checks API](https://docs.github.com/en/rest/checks/runs) (so basically all github-actions tests). `get_combined_status` only takes into account the former, unfortunately

bseverino: [commit statuses](https://docs.github.com/en/rest/commits/statuses) are generated by external services (e.g. codecov), while check runs use the [GH Checks API](https://docs.github.com/en/rest/checks/runs) (so basically all github-actions tests). `get_combined_status` only takes into account the former, unfortunately

bseverino:
> just for my understanding, what is the new check for mergeable_state? where did it come from?

@srwang it was added in a recent commit to fix the job on edx-platform: https://github.com/openedx/ecommerce-scripts/commit/7e8ed6414a3de4123703442a77002e1ede44a2f0

transifex/utils.py
bseverino: on that note, I think our translation PRs were merging without waiting for CI before 😶

noisecapella (approved):

transifex/utils.py
```diff
@@ -340,6 +372,49 @@ def merge_pr(self):
         )
         raise RuntimeError('Pending status checks or other failure prevented this PR from being automatically merged')
 
+    def _attempt_merge(self):
+        try:
+            self.pr.merge(merge_method=self.merge_method)
+            logger.info('Merged [%s/#%d].', self.name, self.pr.number)
+            return True
+        except GithubException as e:
+            logger.error(
```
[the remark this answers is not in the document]
bseverino: can't hurt!

PR #201, opened 2023-03-30
fix: avoid use of use_2to3 in PyGithub setup.py for geoip job

matthugs:
error message like 
Screenshot 2023-03-30 at 11 33 02 AM

which from https://github.com/PyGithub/PyGithub/commit/fec6034a07677b9978a605dc72ab601f6cd1c1b3#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L112 was addressed in 1.44.1

the feature in question is a setup.py which is effectively not maintained for backwards compatibility

### DIFF
--- a/geoipupdate/requirements.txt
+++ b/geoipupdate/requirements.txt
@@ -1,4 +1,4 @@
-PyGithub==1.43.3
+PyGithub==1.44.1
 urllib3==1.24.1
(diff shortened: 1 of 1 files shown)


adzuci (approved):
LGTM

